PR #4, opened 2022-02-22
catch downstream kokkos dependency

Yurlungur:
## PR Summary

As suggested by @mauneyc-LANL this let's the spackage for ports of call catch the kokkos dependency for packages that depend on ports-of-call..

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.


Yurlungur:
Depending on spack version, this may need to be regressed.

Yurlungur:
@mauneyc-LANL approves. Merging.

Yurlungur:
not avail in spack 16.3 which we depend on. Cancelling MR for now.

PR #6, opened 2022-03-31
Sync with spack upstream

Yurlungur:
## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

Same as #5 but with updated docs.


junghans:
Wouldn’t it be better to just drop the spackage here?

Yurlungur:
Ah, I didn't think of that... that's probably fine too.

PR #9, opened 2022-06-21
cmake config guard for submodule mode

mauneyc-LANL:
## PR Summary

If used as in-tree as a submodule downstream, cmake will `include()` a file generated by `export()`, which generates an error (see https://cmake.org/cmake/help/v3.17/policy/CMP0024.html). This includes a guard against invoking the `include()` in the scenario when `ports-of-call` is not the main project. 

This fixes an issue encountered when `singularity-eos` cmake configuration was re-run after the initial cmake generate step. In this case, importing `spiner` submodule would use the prior `export()`-ed `ports-of-callConfig.cmake` that was present in the build directory. Due to CMP0024 (above), this would produce an error, and the build directory would need to be cleaned to produce a usable configuration.

## PR Checklist

- [x] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-callConfig.cmake.in
+++ b/ports-of-callConfig.cmake.in
@@ -5,6 +5,8 @@
 
 @PACKAGE_INIT@
 
-include("${CMAKE_CURRENT_LIST_DIR}/@POCLIB@Targets.cmake")
+if(NOT TARGET @POCLIB@ AND NOT @POCLIB@_BINARY_DIR)
+  include("${CMAKE_CURRENT_LIST_DIR}/@POCLIB@Targets.cmake")
+endif()
 
 check_required_components(@POCLIB@)


Yurlungur (approved):
Good catch thanks!

Yurlungur:
@mauneyc-LANL let me know when all tests pass and I will merge. Can't run tests myself at the moment.

mauneyc-LANL:
> @mauneyc-LANL let me know when all tests pass and I will merge. Can't run tests myself at the moment.

I'm not seeing any tests for this code.

Yurlungur:
I thought there was an install test. Well no big deal either way.

Yurlungur:
The install test passed. Merging.

Yurlungur:
No problem.

PR #23, opened 2022-12-09
Updating ports-of-call version in cmake lists.

dholladay00:
## PR Summary

Bumping cmake version number. @mauneyc-LANL is this the sequence:
- 1. Bump cmake version and merge
- 2. tag commit from (1.) to correct version
- 3. update spackage in new commit ? 

Something doesn't feel right about this sequence but it looks like the spackage can't be updated until it has a commit hash for the tag. 

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


Yurlungur (approved):
that procedure sounds right to me, but I defer to @mauneyc-LANL and @rbberger

mauneyc-LANL:
We can update the `package.py` file after the version is available. Updating the upstream spack would be separate

PR #26, opened 2023-01-06
add portable error handling

Yurlungur:
## PR Summary

In [the Vinet PR](https://github.com/lanl/singularity-eos/pull/202) by @aematts in singularity-eos some issues with the current error handling in singularity-eos were noted. I believe a performance-portable error handling treatment should be developed and properly lives in `ports-of-call`. I introduce it here. In particular, I define several flavors of macro:
- `PORTABLE_REQUIRE` which is an assert statement with file and line numbers
- `PORTABLE_REQUIRE_THROWS` as above but throws an exception
- `PORTABLE_ABORT` which is a fail statement with file and line numbers, resolves to backend aborts appropriate
- `PORTABLE_THROW` resolves to throwing an exception
- `PORTABLE_THROW_IFEXCEPT` resolves to either an exception or an abort depending on portability strategy
- `PORTABLE_WARN` which is a warning with file and line numbers 

All the above are no-ops by default, but are enabled for debug builds. I also provide `PORTABLE_ALWAYS_*` macros which always work.

The idea here is by default error handling should be silent unless you're debugging. But if the developer knows they want a message or a failure, they can insert one. For the Vinet PR, initialization errors should probably be handled with `PORTABLE_THROW_IFEXCEPT` for example.

## PR Checklist

- [x] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -84,23 +84,19 @@ line number where the macro was called, enabling easier debugging.
 The following macros are **disabled** automaticaly for production
 builds (e.g., when the ``NDEBUG`` preprocessor macro is defined):
 
-* ``PORTABLE_REQUIRE(condition, message)`` prints an error message and aborts the program (without throwing an exception) if compiled in debug mode and ``condition`` is not satisfied. This macro **works** in device kernels.
-* ``PORTABLE_REQUIRE_THROWS(condition, message)`` raises a runtime exception (annotated with the message) when compiled in debug mode and the condition is not satisfied. This macro **does not** work in device kernels.
-* ``PORTABLE_ABORT(message)`` prints an error message and aborts the program when compiled in debug mode. This macro **works** in device kernels.
-* ``PORTABLE_THROW(message)`` prints an error message and raises a runtime exception when compiled in debug mode. This macro **does not** work in device kernels.
-* ``PORTABLE_WARN(message)`` prints a warning message if compiled in debug mode. This macro **works** in device kernels.
-* ``PORTABLE_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. As with the above portable, this macro is **disabled in production.** This macro **works** in device kernels.
+* ``PORTABLE_REQUIRE(condition, message)`` prints an error message and aborts the program (without throwing an exception) if compiled in debug mode and ``condition`` is not satisfied.
+* ``PORTABLE_ABORT(message)`` prints an error message and aborts the program when compiled in debug mode.
+* ``PORTABLE_WARN(message)`` prints a warning message if compiled in debug mode.
+* ``PORTABLE_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. This macro is disabled in production.
 
 Each of the above macros is **disabled** and becomes a no-op for most builds and only enabled for ``Debug`` builds. However, for each of the above macros there is an equivalent ``PORTABLE_ALWAYS_*`` macro, which **always** functions and is **never** a no-op:
 
-* ``PORTABLE_ALWAYS_REQUIRE(condition, message)`` prints an error message and aborts the program (without throwing an exception) if ``condition`` is not satisfied. This macro **works** in device kernels.
-* ``PORTABLE_ALWAYS_REQUIRE_THROWS(condition, message)`` raises a runtime exception (annotated with the message) when the condition is not satisfied. This macro **does not** work in device kernels.
-* ``PORTABLE_ALWAYS_ABORT(message)`` prints an error message and aborts the program. This macro **works** in device kernels.
-* ``PORTABLE_ALWAYS_THROW(message)`` prints an error message and raises a runtime exception. This macro **does not** work in device kernels.
-* ``PORTABLE_ALWAYS_WARN(message)`` prints a warning message. This macro **works** in device kernels.
-* ``PORTABLE_ALWAYS_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. This macro **works** in device kernels.
+* ``PORTABLE_ALWAYS_REQUIRE(condition, message)`` prints an error message and aborts the program (without throwing an exception) if ``condition`` is not satisfied.
+* ``PORTABLE_ALWAYS_ABORT(message)`` prints an error message and aborts the program.
+* ``PORTABLE_ALWAYS_WARN(message)`` prints a warning message.
+* ``PORTABLE_ALWAYS_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception.
 
-Please note that none of these functions are thread or MPI aware. Iin a parallel program, the same message may be called **many times**. Therefore caution should be used with this machinery and you may wish to hide these macros in if statements, for example,
+Please note that none of these functions are thread or MPI aware. In a parallel program, the same message may be called **many times**. Therefore caution should be used with this machinery and you may wish to hide these macros in if statements, for example,
 
 .. code-block::
 
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -89,7 +89,7 @@ builds (e.g., when the ``NDEBUG`` preprocessor macro is defined):
 * ``PORTABLE_ABORT(message)`` prints an error message and aborts the program when compiled in debug mode. This macro **works** in device kernels.
 * ``PORTABLE_THROW(message)`` prints an error message and raises a runtime exception when compiled in debug mode. This macro **does not** work in device kernels.
 * ``PORTABLE_WARN(message)`` prints a warning message if compiled in debug mode. This macro **works** in device kernels.
-* ``PORTABLE_THROW_IFEXCEPT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. As with the above portable, this macro is **disabled in production.** This macro **works** in device kernels.
+* ``PORTABLE_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. As with the above portable, this macro is **disabled in production.** This macro **works** in device kernels.
 
 Each of the above macros is **disabled** and becomes a no-op for most builds and only enabled for ``Debug`` builds. However, for each of the above macros there is an equivalent ``PORTABLE_ALWAYS_*`` macro, which **always** functions and is **never** a no-op:
 
@@ -98,7 +98,7 @@ Each of the above macros is **disabled** and becomes a no-op for most builds and
 * ``PORTABLE_ALWAYS_ABORT(message)`` prints an error message and aborts the program. This macro **works** in device kernels.
 * ``PORTABLE_ALWAYS_THROW(message)`` prints an error message and raises a runtime exception. This macro **does not** work in device kernels.
 * ``PORTABLE_ALWAYS_WARN(message)`` prints a warning message. This macro **works** in device kernels.
-* ``PORTABLE_ALWAYS_THROW_IFEXCEPT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. This macro **works** in device kernels.
+* ``PORTABLE_ALWAYS_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. This macro **works** in device kernels.
 
 Please note that none of these functions are thread or MPI aware. Iin a parallel program, the same message may be called **many times**. Therefore caution should be used with this machinery and you may wish to hide these macros in if statements, for example,
 
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -96,6 +96,12 @@ Each of the above macros is **disabled** and becomes a no-op for most builds and
 * ``PORTABLE_ALWAYS_WARN(message)`` prints a warning message.
 * ``PORTABLE_ALWAYS_THROW_OR_ABORT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception.
 
+Additionally the macro
+
+* ``PORTABLE_ERROR_MESSAGE(message, output)`` fills an output ``char*`` with a useful error message containing the filename and line number where the macro is called. Note there is no bounds checking so you **must** provide the macro with a sufficiently large ``char*`` array.
+
+The ``message`` parameter in the above macros can be ``char*`` arrays and string literals on device and additionally accepts ``std::string`` and ``std::stringstream`` on host.
+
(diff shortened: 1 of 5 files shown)


ports-of-call/portable_errors.hpp
```diff
@@ -0,0 +1,198 @@
...
+#define PORTABLE_ALWAYS_WARN(message)                                          \
+  PortsOfCall::ErrorChecking::warn(message, __FILE__, __LINE__);
+
+#ifdef PORTABILITY_STRATEGY_NONE
+#define PORTABLE_ALWAYS_THROW_IFEXCEPT(message) PORTABLE_ALWAYS_THROW(message)
+#else
+#define PORTABLE_ALWAYS_THROW_IFEXCEPT(message) PORTABLE_ALWYAS_ABORT(message)
+#endif // PORTABILITY_STRATEGY_NONE
+
+#fdef NDEBUG
```
mauneyc-LANL, line 57: probably `#ifdef`

doc/sphinx/src/using.rst
```diff
@@ -73,7 +73,42 @@ limited. The syntax is:
...
+``portable_errors.hpp`` provides error handling that works with
+different portability backends, such as `Kokkos`. We provide several
+useful macros. All the macros in this file will print the file and
+line number where the macro was called, enabling easier debugging.
+
+The following macros are **disabled** automaticaly for production
+builds (e.g., when the ``NDEBUG`` preprocessor macro is defined):
+
+* ``PORTABLE_REQUIRE(condition, message)`` prints an error message and aborts the program (without throwing an exception) if compiled in debug mode and ``condition`` is not satisfied. This macro **works** in device kernels.
+* ``PORTABLE_REQUIRE_THROWS(condition, message)`` raises a runtime exception (annotated with the message) when compiled in debug mode and the condition is not satisfied. This macro **does not** work in device kernels.
```
jhp-lanl, line 88: Given that many of these _do not_ work on device, what then is the definition of "portable"?

ports-of-call/portable_errors.hpp
```diff
@@ -0,0 +1,198 @@
...
+#define PORTABLE_THROW(message) PORTABLE_ALWAYS_THROW(message)
+#define PORTABLE_WARN(message) PORTABLE_ALWAYS_WARN(message)
+#define PORTABLE_THROW_IFEXCEPT(message) PORTABLE_ALWAYS_THROW_IFEXCEPT(message)
+#endif // NDEBUG
+
+namespace PortsOfCall {
+namespace ErrorChecking {
+namespace impl {
+[[noreturn]] KOKKOS_INLINE_FUNCTION void abort(const char *const message) {
+#ifdef PORTABILITY_STRATEGY_KOKKOS
```
jhp-lanl, line 79: Shouldn't the `KOKKOS_INLINE_FUNCTION` be behind the `ifdef`?

jhp-lanl (commented):
This will be very helpful I think. But rather than some of these just not working on device, should they instead do nothing? My issue is that if a "portable" implementation doesn't work on device, it begs the question as to whether it's truly portable anymore.

ports-of-call/portable_errors.hpp
jhp-lanl: Or should it use `PORTABLE_FUNCTION`?

ports-of-call/portable_errors.hpp
Yurlungur: thanks for the catch

Yurlungur:
> But rather than some of these just not working on device, should they instead do nothing? My issue is that if a "portable" implementation doesn't work on device, it begs the question as to whether it's truly portable anymore.

That's a fair point. This was written this way to support exception handling on host, when that's desirable. For example, you can imagine some initialization function that uses C++ std exception logic. Basically I wanted to proved that, also for testing. For example the catch machinery Anne wants use for Vinet catches exceptions.

doc/sphinx/src/using.rst
```diff
@@ -73,7 +73,42 @@ limited. The syntax is:
...
+* ``PORTABLE_ABORT(message)`` prints an error message and aborts the program when compiled in debug mode. This macro **works** in device kernels.
+* ``PORTABLE_THROW(message)`` prints an error message and raises a runtime exception when compiled in debug mode. This macro **does not** work in device kernels.
+* ``PORTABLE_WARN(message)`` prints a warning message if compiled in debug mode. This macro **works** in device kernels.
+* ``PORTABLE_THROW_IFEXCEPT(message)`` prints an error message and then raises a runtime error if ``PORTABILITY_STRATEGY`` is ``NONE`` and otherwise aborts the program without an exception. As with the above portable, this macro is **disabled in production.** This macro **works** in device kernels.
```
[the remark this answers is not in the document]
Yurlungur: Yeah that makes sense. I'll change the name.

doc/sphinx/src/using.rst
Yurlungur: In this case I'm just namespacing, because the macros can't live in a `PortsOfCall` namespace, so really `PORTABLE` here is `PORTS_OF_CALL`

I think it makes sense to provide a unified interface though with the truly portable versions and the not-so-portable versions that use exception handling. The issue is exception handling simply does not work on device, but we may still want to provide it. Here by exception handling I mean specifically the machinery provided by `stdexcept` which provides things like
```C++
try {
  // some code
  if (condition) {
    PORTABLE_THROW(message)
  } catch (const std::runtime_error &error) {
    // some code that recovers from the error gracefully
  }
}
```
what `ports-of-call` is providing on top of that is a filename and line number, as well as `THROW_OR_ABORT` which gracefully switches between the exception model and the "print an error and die" model.

ports-of-call/portable_errors.hpp
Yurlungur: It should just be `PORTABLE_INLINE_FUNCTION` my bad.

doc/sphinx/src/using.rst
Yurlungur: I suppose we could not provide `REQUIRE_THROWS` or `THROWS` and only provide `THROW_OR_ABORT`. But I think sometimes we actually do want to just throw.

chadmeyer: > This will be very helpful I think. But rather than some of these just not working on device, should they instead do nothing? My issue is that if a "portable" implementation doesn't work on device, it begs the question as to whether it's truly portable anymore.

I recognize that having a *universal* error handling mechanism is probably a pipe dream, but is it not possible to have something that could yet function on device and do something?  I seem to think that we've had things that can abort a kernel and then set some error state (which may not be seen until the next kernel launch) but might be limited to CUDA or something like that (which, to be clear, we should support).  I'm certainly no expert on error handling!  I think we should be careful about putting together something that could easily be snuck in to some inappropriate parts of library codes that would then make for some downstream headaches; at the very least if the macros have to expand to nothing on device that would be better than having them expand to something with undefined behavior.

Yurlungur: It is possible to have something that throws an error on host and aborts on device. I could remove the functions that only work on host if that's how people feel, and leave things that are all truly portable. The only disadvantage of this is that we will be excluding any use of C++ exceptions if anyone cares about that. But I suppose if someone wants to use that it can be on them.

chadmeyer: My main concern is the case where a downstream library puts in this kind of functionality to ensure certain behavior, but in reality this is device code (at least sometimes) and the behavior becomes either undefined or wrong/unexpected.  I can imagine it being a very hard bug to track down if there is some kind of silent-ish abort of a device-side constructor or something similar but the host side is never the wiser.  

Maybe I'm being a little too paranoid about this here?

Yurlungur: so I *think* the functionality that doesn't work on device will fail to compile. But I should double check that. We *can* simply not provide the functionality that might fail. The reason I'm reluctant to do that is I feel that there are times when C++ exceptions provide a useful ability on host.

The aborts that *do* run on device will all run on host or device and will be loud.

ports-of-call/portable_errors.hpp
```diff
@@ -0,0 +1,198 @@
...
+                               const char *const filename,
+                               int const linenumber) {
+  abort(message.str().c_str(), filename, linenumber);
+}
+
+[[noreturn]] inline void abort(std::string const &message,
+                               const char *const filename,
+                               int const linenumber) {
+  abort(message.str().c_str(), filename, linenumber);
+}
```
jhp-lanl, line 153: Why not template on the type of `msg` rather than overloading on explicit types?

doc/sphinx/src/using.rst
jhp-lanl: > In this case I'm just namespacing, because the macros can't live in a `PortsOfCall` namespace, so really `PORTABLE` here is `PORTS_OF_CALL`
> 
> I think it makes sense to provide a unified interface though with the truly portable versions and the not-so-portable versions that use exception handling. The issue is exception handling simply does not work on device, but we may still want to provide it. Here by exception handling I mean specifically the machinery provided by `stdexcept` which provides things like
> 
> ```c++
> try {
>   // some code
>   if (condition) {
>     PORTABLE_THROW(message)
>   } catch (const std::runtime_error &error) {
>     // some code that recovers from the error gracefully
>   }
> }
> ```
> 
> what `ports-of-call` is providing on top of that is a filename and line number, as well as `THROW_OR_ABORT` which gracefully switches between the exception model and the "print an error and die" model.

But what is the true _utililty_ of `PORTABLE_THROW`? What if I modify the above code (I assume it should be try/catch rather than try{if/catch}, but my C++ knowledge is limited) to be
```c++
try {
   // some code
   if (condition) {
      const char* const message = PORTABLE_ERR_MSG(message)
      std::runtime_error(msg)
   }
} catch (const std::runtime_error &error) {
   // some code that recovers from the error gracefully
}
```
If the utility is in the creation of the error message, should that then be the functionality that is exposed?

Yurlungur: Hmm. Yeah that's a good point. I like the idea of exposing the error message generation and removing the throw machinery... with the exception of the macro that either throws or aborts depending.

Yurlungur: Ok I've made the change. Here's how it works now:
- `PORTABLE_THROW` is gone.
- `PORTABLE_THROW_OR_ABORT` still exists and is the preferred way to perform checks that you would like to throw an exception but want to work on device.
- I have added `PORTABLE_ERROR_MESSAGE` . It can't work as cleanly as your suggested code above @jhp-lanl because that kind of assignment leaks memory, so you have to fill an array that is passed in as an argument. It also has to be a `char*` to work on device, which means the user is responsible for allocating enough memory to support their string on device without a segfault. But I think that's fine.

jhp-lanl: I was thinking about this yesterday but forgot to chime in. Would a compile time error message length solve the memory leak issue? Could the length default to something and be set via CMake?

BUUUUUT, I don't feel strongly. Simply printing something and then raising an exception is also fine with me.

jhp-lanl (approved):
Provisionally approving since I think this addresses my comments. I'd like @dholladay00 or @chadmeyer to give final approval though

doc/sphinx/src/using.rst
Yurlungur: > I was thinking about this yesterday but forgot to chime in. Would a compile time error message length solve the memory leak issue? Could the length default to something and be set via CMake?

That would allow us to fix length ahead of time, which would avoid the segfault thing, but it wouldn't allow us to return a pointer. We'd still have to fill an array that's passed to us. I think the current design is fine, though I'm not opposed to a compile-time string length if people feel strongly.

Yurlungur:
Yep I'll wait to hear from @chadmeyer and @dholladay00 before merging.

Yurlungur:
@dholladay00 and @chadmeyer what are your thoughts here?

mauneyc-LANL:
Some documentation in the code wouldn't hurt, though I leave it to your discretion.

Yurlungur:
> Some documentation in the code wouldn't hurt, though I leave it to your discretion.

I don't mind adding more documentation. I didn't add any here because I didn't see what would be a useful addition in this case. What would you find helpful @mauneyc-LANL ?

mauneyc-LANL:
> > Some documentation in the code wouldn't hurt, though I leave it to your discretion.
> 
> I don't mind adding more documentation. I didn't add any here because I didn't see what would be a useful addition in this case. What would you find helpful @mauneyc-LANL ?

Really just a bit of comments to help "paragraph" the code a little and make it easier to parse.

```c++
/***
Used when conditions are encountered where execution could not complete. 
Takes an `input_message`, `filename`, and `line_no`  describing the error, and 
wraps it into formatted output `output_message`.
***/
PORTABLE_INLINE_FUNCTION
void error_msg(const char *const input_message, const char *const filename,
               int const linenumber, char *output_message) {
  std::sprintf(output_message,
               "### ERROR\n  Message:     %s\n  File:        %s\n  Line number: %i\n",
               input_message, filename, linenumber);
}
```
(probably something less illiterate)

ports-of-call/portable_errors.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+inline void error_msg(std::stringstream const &input_message, const char *const filename,
+                      int const linenumber, char *output_message) {
+  error_msg(input_message.str().c_str(), filename, linenumber, output_message);
+}
+
+inline void error_msg(std::string const &message, const char *const filename,
+                        int const linenumber) {
+  error_msg(input_message.c_str(), filename, linenumber, output_message);
```
mauneyc-LANL, line 179: If I follow, you're missing the `output_message` parameter here.

Yurlungur:
> > > Some documentation in the code wouldn't hurt, though I leave it to your discretion.
> > 
> > 
> > I don't mind adding more documentation. I didn't add any here because I didn't see what would be a useful addition in this case. What would you find helpful @mauneyc-LANL ?
> 
> Really just a bit of comments to help "paragraph" the code a little and make it easier to parse.
> 
> ```c++
> /***
> Used when conditions are encountered where execution could not complete. 
> Takes an `input_message`, `filename`, and `line_no`  describing the error, and 
> wraps it into formatted output `output_message`.
> ***/
> PORTABLE_INLINE_FUNCTION
> void error_msg(const char *const input_message, const char *const filename,
>                int const linenumber, char *output_message) {
>   std::sprintf(output_message,
>                "### ERROR\n  Message:     %s\n  File:        %s\n  Line number: %i\n",
>                input_message, filename, linenumber);
> }
> ```
> 
> (probably something less illiterate)

Sure I'll add a little bit.

Yurlungur:
@mauneyc-LANL take a look now. Does this help? I tired to describe the intent of the design as well... e.g., why there are both macros and functions.

mauneyc-LANL:
> @mauneyc-LANL take a look now. Does this help? I tried to describe the intent of the design as well... e.g., why there are both macros and functions.

looks good

jhp-lanl:
@Yurlungur is this ready now?

ports-of-call/portable_errors.hpp
```diff
@@ -0,0 +1,209 @@
...
+// assert (the assertion is applied above in the macro) with file name
+// and line number. Then aborts.
+[[noreturn]] PORTABLE_INLINE_FUNCTION void require(const char *const condition,
+                                                   const char *const message,
+                                                   const char *const filename,
+                                                   int const linenumber) {
+  std::printf("### ERROR\n  Condition:   %s\n  Message:     %s\n  File:        "
+              "%s\n  Line number: %i\n",
+              condition, message, filename, linenumber);
+  impl::abort(message);
```
dholladay00, line 117: won't this double print the message? same for below

Yurlungur: yes it does---personally I think this is fine, but if it bothers you I can tweak to avoid that

Yurlungur:
> @Yurlungur is this ready now?

as soon as I get OK from @dholladay00 and/or @chadmeyer

ports-of-call/portable_errors.hpp
dholladay00: If this happens to get used at a low level, it could get very annoying (ie **twice** per material, per iteration (if pte solve), per cell, per rank, per time step). Could you just pass in an empty string?

Yurlungur: Yeah I'll do that.

Yurlungur:
@dholladay00 how's it look now?

dholladay00 (approved):

PR #30, opened 2023-02-16
Change behavior of `require` to match documentation

jhp-lanl:
## PR Summary
If we 'require' something, then the _failed_ condition should trigger the error message, not the success. This small change brings the code in line with the documentation and my understanding of the intent of the `REQUIRE` macros.

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.
- N/A If preparing for a new release, update the version in cmake.

### DIFF
--- a/ports-of-call/portable_errors.hpp
+++ b/ports-of-call/portable_errors.hpp
@@ -111,7 +111,7 @@ namespace impl {
                                                    const char *const message,
                                                    const char *const filename,
                                                    int const linenumber) {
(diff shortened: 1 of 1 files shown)


jhp-lanl:
@Yurlungur let me know if I'm missing something, but it looks like `require()` should only have an error when the condition is `false` rather than `true`.

Yurlungur (approved):
Oops thanks for the catch!

PR #31, opened 2023-02-18
Remove [[noreturn]] from 'require' since it should return under most circumstances

jhp-lanl:
## PR Summary

The `require()` function should usually _not_ return since we would hope the condition is `false`. The `[[noreturn]]` decorator applied to a returning function is undefined behavior. With gcc 10, it looks like the compiler uses `[[noreturn]]` to just give up on keeping track of the stack pointer which leads to a segfault.

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.

### DIFF
--- a/ports-of-call/portable_errors.hpp
+++ b/ports-of-call/portable_errors.hpp
@@ -107,10 +107,10 @@ namespace impl {
 // assert (the assertion is applied above in the macro) with file name
 // and line number. Then aborts.
 PORTABLE_INLINE_FUNCTION void require(bool condition_bool,
-                                                   const char *const condition,
-                                                   const char *const message,
-                                                   const char *const filename,
-                                                   int const linenumber) {
+                                      const char *const condition,
+                                      const char *const message,
+                                      const char *const filename,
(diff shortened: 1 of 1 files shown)


Yurlungur:
Thanks for fixing this @jhp-lanl . If we're making this change we may also wish to remove `[[noreturn]]` from `abort` and remove the `while (true) {}` loop in there, which is supposed to make the compiler not complain about `[[noreturn]]` but doesn't seem to work. But that can be for a future issue.

Yurlungur (approved):

jhp-lanl:
> Thanks for fixing this @jhp-lanl . If we're making this change we may also wish to remove `[[noreturn]]` from `abort` and remove the `while (true) {}` loop in there, which is supposed to make the compiler not complain about `[[noreturn]]` but doesn't seem to work. But that can be for a future issue.

I'm less concerned about this since I've seen infinite loops used for asynchronous error handling in other places. Basically the loop can exist to give all processes something to do until the one that does all the aborting can actually abort.

PR #32, opened 2023-03-02
Add core for build tests

mauneyc-LANL:
## PR Summary

Currently the only test in `ports-of-call` is the install test. This PR does a minimal build configuration for inline tests.

I have included a test from `spiner` that tests `PortableMDArray`, but that is just so there is something to test. It appears there are several tests in `spiner` that should properly be in `ports-of-call`.

New CMake options:

- `-DBUILD_TESTING=[ON/OFF]` for triggering the build of the test exe(s)
- `-DPORTS_OF_CALL_TEST_KOKKOS=[ON/OFF]` for configuring for testing with `Kokkos`

### Dependencies

I'm taking a "less-is-more" approach here; if you want to test, you need `Catch2` available - diddo with `Kokkos`. Further, I've deliberately left *out* an option for Kokkos CUDA; If you want that, then install Kokkos with CUDA enabled.

### CI ?

Should the git[hub/lab] CI begin running these tests? Or should that be a separate PR?

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ target_compile_features(${POCLIB}
 # ----------------------------------------
 if(PORTS_OF_CALL_BUILD_TESTING)
   message(STATUS "Configuring tests")
+  enable_testing()
   add_subdirectory(test)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,9 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
     "Please make a build subdirectory. Feel free to remove CMakeCache.txt and CMakeFiles.")
 endif()
 
+include(CTest) # brings in option `BUILD_TESTING`
+
+option (PORTS_OF_CALL_TEST_USE_KOKKOS "Use kokkos offloading for testing")
 
 # CONFIGURATION LOGIC
 # ----------------------------------------
@@ -62,6 +65,13 @@ target_compile_features(${POCLIB}
   INTERFACE
     cxx_std_14
 )
+
+# TESTING
+# ----------------------------------------
+if(BUILD_TESTING)
+  message(STATUS "Configuring tests")
+  add_subdirectory(test)
+endif()
 
 # INSTALL & EXPORT
 # ----------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,9 +31,7 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
     "Please make a build subdirectory. Feel free to remove CMakeCache.txt and CMakeFiles.")
 endif()
 
-include(CTest) # brings in option `BUILD_TESTING`
-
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+# compile/link options for the test
+add_library(portsofcall_iface INTERFACE)
+# instructions for local content
+# https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_SOURCE_DIR_%3CuppercaseName%3E
+# TLDR: set FETCHCONTENT_SOURCE_DIR_KOKKOS=path/to/local/checkout
+if (PORTS_OF_CALL_TEST_USE_KOKKOS)
+  find_package(Kokkos REQUIRED)
+  target_link_libraries(portsofcall_iface INTERFACE Kokkos::kokkos)
+  target_compile_definitions(portsofcall_iface INTERFACE PORTABILITY_STRATEGY_KOKKOS)
+endif()
```
Yurlungur, line 26: This should be inside a kokkos include guard.

CMakeLists.txt
```diff
@@ -63,6 +66,13 @@ target_compile_features(${POCLIB}
     cxx_std_14
 )
 
+# TESTING
+# ----------------------------------------
+if(BUILD_TESTING)
+  message(STATUS "Configuring tests")
+  add_subdirectory(test)
+endif()
```
Yurlungur, line 74: I would suggest we don't use `BUILD_TESTING` as this tends to turn on *all* the testing of all dependencies. If ports is built in-tree, this could be annoying. I would prefer a `TEST_PORTS_OF_CALL` or `PORTS_BUILD_TESTING` or something similar.

Yurlungur (approved):
Thanks for adding this @mauneyc-LANL . I think CI can be a separate PR but it's up to you and @dholladay00

test/CMakeLists.txt
dholladay00: Could this be based on portability strategy?

Yurlungur: it could I still would like an include guard here. In case this somehow gets called in an in-tree build. Just good practice.

mauneyc-LANL: Please see latest commit for updates with these suggestions

dholladay00:
as long as @Yurlungur is good with the option name this looks g2g.

dholladay00 (approved):

test/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+# prepare derivative works, distribute copies to the public, perform
+# publicly and display publicly, and to permit others to do so.
+
+find_package(Catch2 REQUIRED)
+
+# this interface target is to collect
+# compile/link options for the test
+add_library(portsofcall_iface INTERFACE)
+if (PORTABILITY_STRATEGY_KOKKOS)
+  if(NOT TARGET Kokkos::kokkos)
```
Yurlungur, line 19: :+1:

Yurlungur (approved):
LGTM let's get it in

PR #33, opened 2023-03-31
Draft: Add portable query for exec space

mauneyc-LANL:
## PR Summary

Adds `portableExecIsHost()` function that returns `true` if the execution space is on host, `false` if on device.

For `PORTABILITY_STRATEGY_KOKKOS`, compares `Kokkos::DefaultExecutionSpace` to `Kokkos::HostSpace::execution_space`. This can differentiate between `openmp` and `cuda`, for instance

For the explicit `PORTABILITY_STRATEGY_*`, just return `true` or `false` as expected.

This maybe useful e.g. https://github.com/lanl/spiner/pull/64

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.
- [x] If preparing for a new release, update the version in cmake.

### DIFF
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -83,6 +83,14 @@ with `to` being the target location, from being the source location, and size_by
 the size of the transfer in bytes. This has implemenatations for kokkos and none 
 portability strategies.
 
+It may be useful to query the execution space, for example to know where memory needs to be copied.
+To this end, a compile-time constant boolean can be queried:
+
+.. cpp:var:: PortsOfCall::EXECUTION_IS_HOST
+
+which is `true` if the host execution space can trivially access device memory space. For example,
+for `PORTABILITY_STRATEGY_CUDA`, `PortsOfCall::EXECUTION_IS_HOST == false`.
+
 portable_errors.hpp
 ^^^^^^^^^^^^^^^^^^^^
 
--- a/ports-of-call/portability.hpp
+++ b/ports-of-call/portability.hpp
@@ -91,6 +91,18 @@ typedef float Real;
 typedef double Real;
 #endif
 
+bool portableExecIsHost()
+{
+#ifdef PORTABILITY_STRATEGY_KOKKOS
+  // check if default exec space is the same as the host exec space
+  return std::is_same<Kokkos::DefaultExecutionSpace,Kokkos::HostSpace::execution_space>::value;
+#elif PORTABILITY_STRATEGY_CUDA
+  return false;
+#else
+  return true;
+#endif
+}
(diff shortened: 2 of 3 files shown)


dholladay00:
@mauneyc-LANL I think using something similar to what I had is a bit more robust, for example does that do the right thing if OpenMP is the default memory space, I think this code below gets at the heart of the matter (do I need to copy memory somewhere else), or possibly we need this and other variables for different use cases. 

```c++
     using HS = Kokkos::HostSpace;
     using DMS = Kokkos::DefaultExecutionSpace::memory_space;
     constexpr const bool execution_is_host{
         Kokkos::SpaceAccessibility<DMS, HS>::accessible};
```

Yurlungur (changes requested):
Thanks for adding this, @mauneyc-LANL ! I think this is a good solution. @dholladay00 and I were talking about this before we saw your MR and were thinking of exposing a function call, but you're right there's no need to. All information is available at compile time.

That said, should we make it a function call anyway? I'm thinking about systems like openmp target. Is that introspection available at compile time? If it is, let's go with this solution. If it's not, then we may need to think.

Either way it would probably also be good to make this parameter all caps, rather than lower-case, to indicate that it's a constant.

One more "food for thought" question: Should we hide this in a `PortsOfCall` or `Portable` namespace?

mauneyc-LANL:
@Yurlungur My initial commit was a function, but that was changed to a constant; I have no preference, though the constant should be namespaced - just let me know what namespace you want here.

The OpenMP specification has stuff to query the memory space, and Kokkos probably has some code committed to do this.

chadmeyer:
I agree with the sentiment of @Yurlungur that this should be compile time known so it should be a compile-time-evaluatable quantity -- whether a `constexpr` variable or `constexpr` function or whatnot.  I'm not sure I know enough to weigh in as to which option is going to be better behaved.

mauneyc-LANL:
@Yurlungur see 8204a93 for suggested changes

Yurlungur:
LGTM thanks @mauneyc-LANL . One last request: Can you add one line to the docs mentioning this constant exists?

mauneyc-LANL:
@Yurlungur see e82175b

Yurlungur (approved):
LGTM. Assuming tests pass and @dholladay00 is happy let's get this in.

dholladay00 (approved):
I approve conditionally based on tests passing (as stated in previous comment with none and kokkos portability strategies).

mauneyc-LANL:
Testing passes with kokkos

dholladay00:
I'm going to merge this in.

PR #35, opened 2023-04-07
removed inline, icpc complains

mauneyc-LANL:
## PR Summary

Remove the `inline` keyword from `EXECUTION_IS_HOST` var; `icpc` doesn't like this

## PR Checklist

- [x] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


Yurlungur (approved):

dholladay00:
This is breaking some things for some projects. Could we bump the patch number and release with the inline fix @mauneyc-LANL ?

PR #39, opened 2023-10-02
Templated MD arrays

mauneyc-LANL:
## PR Summary

This is a prototype for "templated MD", that is the reduction of code of the type

```c++
PORTABLE_FUNCTION PortableMDArray(T *data, int nx1) noexcept
      : pdata_(data), nx1_(nx1), nx2_(1), nx3_(1), nx4_(1), nx5_(1), nx6_(1) {}
  PORTABLE_FUNCTION
  PortableMDArray(T *data, int nx2, int nx1) noexcept
      : pdata_(data), nx1_(nx1), nx2_(nx2), nx3_(1), nx4_(1), nx5_(1), nx6_(1) {
  }
  PORTABLE_FUNCTION
  PortableMDArray(T *data, int nx3, int nx2, int nx1) noexcept
      : pdata_(data), nx1_(nx1), nx2_(nx2), nx3_(nx3), nx4_(1), nx5_(1),
        nx6_(1) {}
```

To

```c++
  template <typename... NXs>
  PORTABLE_FUNCTION PortableMDArray(T *p, NXs... nxs) noexcept
      : PortableMDArray(p, nx_arr(nxs...), sizeof...(NXs)) {}
```

The single-value dimension variables `int nxN_` have been replaced with a staticly sized container `std::array<size_t, MAXDIM> nxs_`. Also added `PortableMDArray::rank_` member to avoid some reevaluations, though I'm not sure it's necessary.

Currently WIP, so documentation is wanting and a few function names/namespaces are silly/inconsistent/gibberish. I know there can be some reluctance about this coding style, so I wanted to get a "at least compiles" version up and gauge interest in iterating on it. (Actually passes a few tests! but haven't tried `spiner` with this yet).

### It was working fine, why change it?

- Update code to more modern style
- Code is easier to maintain and diagnose
- Allows for forward planning when C++ standards are updated
- No change in down-facing API
- For developers, the code is easier to read/glance

### What are the downsides?

- Some extra boilerplate to handle C++14's constrains on variadic programming (e.g. no `std::apply()`, no `std::tuple`, no fold expressions). C++14 can be "recursion heavy" in this regard, which may degrade performance.
- API not "explicit" w.r.t. index ordering
- (possibly) Longer compile times
- For downstream users, code or errors passing through/originating in expanded template types can be agitating

Any suggestions, comments, questions welcome! @jonahm-LANL @chadmeyer @dholladay00 @jhp-lanl @jdolence 

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


ports-of-call/portable_arrays.hpp
```diff
@@ -26,73 +26,142 @@
...
+constexpr std::size_t MAXDIM = 6;
+// array type of dimensions/strides
+using narr = std::array<std::size_t, MAXDIM>;
+
+namespace detail {
+
+// set_value end-case
+template <std::size_t Ind, typename NX>
+constexpr void set_value(narr &ndat, NX value) {
+  ndat[Ind] = value;
```
dholladay00, line 48: is `operator[]` of `std::array` marked  `constexpr`? I thought this was something that preventing more widespread usage.

ports-of-call/portable_arrays.hpp
```diff
@@ -26,73 +26,142 @@
...
+constexpr void set_values(narr &ndat, NX value) {
+  set_value<Ind>(ndat, value);
+}
+
+// set_values general case, set `Ind` position of array (tracking from RIGHT)
+// with head of size list `NX`
+template <std::size_t Ind, typename NX, typename... NXs>
+constexpr void set_values(narr &ndat, NX value, NXs... nxs) {
+  set_value<Ind>(ndat, value);
+  set_values<Ind - 1>(ndat, nxs...);
```
dholladay00, line 62: does this not need a 0 partial specialization to terminate the recursion?

mauneyc-LANL: The recursion is limited by running out of variadic parameters, or rather `select_value` is the 0 specialization

```c++
set_values(arr, n3, n2,n1) --> set_values(arr, n2, n1) --> set_values(arr, n1) -->set_value(arr, n1)
```

ports-of-call/portable_arrays.hpp
mauneyc-LANL: Sorry I intended to replace all these with appropriate `PORTABLE_` s but missed a few

mauneyc-LANL: Oh, you probably meant something else. We can use a raw array, if `std::array` is messing with things

ports-of-call/portable_arrays.hpp
```diff
@@ -113,48 +182,21 @@ class PortableMDArray {
     return -1;
   }
 
-  // a function to get the total size of the array
   PORTABLE_FORCEINLINE_FUNCTION int GetSize() const {
-    return nx1_ * nx2_ * nx3_ * nx4_ * nx5_ * nx6_;
+    return std::accumulate(nxs_.cbegin(), nxs_.cend(), 1,
```
dholladay00, line 189: Have you tested this on GPUs? I know Brendan had an issue and we had to write a custom accumulate. One could still use binary operators provided by the STL, but the accumulate itself wouldn't run. Perhaps we did something wrong, but I want to make sure this is still GPU callable.

ports-of-call/portable_arrays.hpp
dholladay00: Thanks for the explanation @mauneyc-LANL I see it now.

ports-of-call/portable_arrays.hpp
Yurlungur: We should check that it runs on device, but I *thought* `std::array` worked with `--expt-relaxed-constexpr` but I'm not sure

ports-of-call/portable_arrays.hpp
Yurlungur: I would also be concerned about `std::accumulate`. Might be better to hardcode this one, or write a a recursive thing ourselves.

ports-of-call/portable_arrays.hpp
```diff
@@ -26,73 +26,145 @@
...
 #include <cstring> // memset()
+#include <functional>
+#include <numeric>
+#include <type_traits>
 #include <utility> // swap()
 
+// maximum number of dimensions
+constexpr std::size_t MAXDIM = 6;
+// array type of dimensions/strides
+using narr = std::array<std::size_t, MAXDIM>;
```
Yurlungur, line 41: don't love that this is lower-case. I'd find this type easier to interpret if it were, e.g.,
```suggestion
using Narr_t = std::array<std::size_t, MAXDIM>;
```

ports-of-call/portable_arrays.hpp
```diff
@@ -26,73 +26,145 @@
...
+  return index;
+}
+
+// compute_index general case, computing slower moving index strides
+template <std::size_t Ind, typename... Tail>
+PORTABLE_INLINE_FUNCTION size_t compute_index(const narr &nd,
+                                              const size_t index,
+                                              const Tail... tail) {
+  return index * nd[Ind] + compute_index<Ind + 1>(nd, tail...);
+}
```
Yurlungur, line 78: This is nice---assuming the array on device issues work.

ports-of-call/portable_arrays.hpp
```diff
@@ -26,73 +26,145 @@
...
+#include <numeric>
+#include <type_traits>
 #include <utility> // swap()
 
+// maximum number of dimensions
+constexpr std::size_t MAXDIM = 6;
+// array type of dimensions/strides
+using narr = std::array<std::size_t, MAXDIM>;
+
+namespace detail {
```
Yurlungur, line 43: It doesn't matter very much, by why `detail` over `impl`?

Yurlungur (commented):
This is a nice cleanup. Before merging, I'd like to:
a) Get a sense of compile time differences
b) Know for sure it all works on device

ports-of-call/portable_arrays.hpp
dholladay00: @mauneyc-LANL you may be able to use `std::get<Ind>(arr) = value;`. But we really need to check and make sure it works on device.

ports-of-call/portable_arrays.hpp
mauneyc-LANL: It "worked" but gave warnings (even with `--expt-relaxed-constexpr`). I rewrote it as an explicit loop.

ports-of-call/portable_arrays.hpp
mauneyc-LANL: Using `--expt-relaxed-constexpr` this runs smoothly using `volta-x86` partition, `kokkos+cuda cuda_arch=70 +wrapper`, `cmake .. -DPORTS_OF_CALL_BUILD_TESTING -DPORTABILITY_STRATEGY_KOKKOS=ON`

ports-of-call/portable_arrays.hpp
mauneyc-LANL: Tests complete on `volta-x86`, though we need to put together some more tests for `ports-of-call`

ports-of-call/portable_arrays.hpp
mauneyc-LANL: Just habit

CMakeLists.txt
```diff
@@ -64,6 +64,14 @@ target_compile_features(${POCLIB}
     cxx_std_14
 )
 
+#NOTE this option is done downstream, maybe poc should control it
+set(with_cxx "$<COMPILE_LANGUAGE:CXX>")
+set(ps_kokkos "$<BOOL:${PORTABILITY_STRATEGY_KOKKOS}>")
+
+target_compile_options(${POCLIB}
+INTERFACE
+$<${with_cxx}:$<${ps_kokkos}:--expt-relaxed-constexpr>>)
```
dholladay00, line 73: In the event of +kokkos~cuda, wouldn't this result in a compile error?

mauneyc-LANL: Looks like it, but POC doesn't have a `USE_CUDA` or similar like `singularity-eos` does, so I need to probe which compiler is being used.

mauneyc-LANL:
I've made some changes to push a lot of the indexing/building functions into the object. I don't know that I love the result, which lead to the elimination of some constexpr initialization (not that it couldnt be done, just saved time to get a working code).

One question to ask is how much time should `PortableMDArray` devote to reconstructing internal layout configuration data - in particular strides. These are the offset widths associated with a particlar dimensional spec, eg.

```
// ordering based on what MDArray is doing
dims = {NT, NZ, NY, NX};
strides = {NZ * NY * NX, NY * NX, NX, 1};
```

The options are:
1. at each `compute_index` call, recompute the strides. This obviously costs some FLOPS but strides are simple to compute and GPU compute is cheap. Further, because the rank is known at this point, the stride data only needs to go up to `rank`, rather than compute/store values up to `MAXDIM`
2. (current commit) at each construction or reshaping, recompute the strides. This is the "natural" path but does introduce more data to move around, to a first approximation doubling the data that `PortableMDArray` needs to move from between contexts. Contra to (1), this array needs `MAXDIM` storage which is likely going to be mostly dead weight.

CMakeLists.txt
dholladay00: This is how I've done it in other projects:

```cmake
  get_target_property(kokkos_link_libs Kokkos::kokkoscore
                      INTERFACE_LINK_LIBRARIES)
  string(REGEX MATCH "CUDA"
         kokkos_has_cuda_libs "${kokkos_link_libs}")
  ...
  if(kokkos_has_cuda_libs)
    do cuda stuff...
  endif()
```

Yurlungur:
> I've made some changes to push a lot of the indexing/building functions into the object. I don't know that I love the result, which lead to the elimination of some constexpr initialization (not that it couldnt be done, just saved time to get a working code).
> 
> One question to ask is how much time should `PortableMDArray` devote to reconstructing internal layout configuration data - in particular strides. These are the offset widths associated with a particlar dimensional spec, eg.
> 
> ```
> // ordering based on what MDArray is doing
> dims = {NT, NZ, NY, NX};
> strides = {NZ * NY * NX, NY * NX, NX, 1};
> ```
> 
> The options are:
> 
>     1. at each `compute_index` call, recompute the strides. This obviously costs some FLOPS but strides are simple to compute and GPU compute is cheap. Further, because the rank is known at this point, the stride data only needs to go up to `rank`, rather than compute/store values up to `MAXDIM`
> 
>     2. (current commit) at each construction or reshaping, recompute the strides. This is the "natural" path but does introduce more data to move around, to a first approximation doubling the data that `PortableMDArray` needs to move from between contexts. Contra to (1), this array needs `MAXDIM` storage which is likely going to be mostly dead weight.

I don't know the answer to this a priori. I'm not married to either option. Which leads to the cleaner code? Probably 2? And do we see a performance hit in a simple test case?

mauneyc-LANL:
> I don't know the answer to this a priori. I'm not married to either option. Which leads to the cleaner code? Probably 2? And do we see a performance hit in a simple test case?

Would be work considering the context that `PortableMDArray` is used in. Does it reshape often, does it get move on/off device frequently, ect. (2.) would be "best" for host, since it avoids recomputation, but (1.) may be better for device since it would minimize data transfer (tho also it's like 48 bytes extra so probably over-optimizing here).

I've included a simple test with the PR. As originally there wasn't very much being tested directly with `poc` we may need to add some more for checking these details. I glanced over spiner tests for ideas but most of what is there looks more numerics based.

Yurlungur:
Yeah fair questions:

> Does it reshape often?

No, almost never.

> does it get move on/off device frequently

Also no.

> (2.) would be "best" for host, since it avoids recomputation, but (1.) may be better for device since it would minimize data transfer (tho also it's like 48 bytes extra so probably over-optimizing here).

Like you said, 48 bytes is basically nothing, so I lean towards (2).

dholladay00:
Previously we recalculated at every access it seems. It seems like the number of adds will be the same and the number of multiplies saved will only be noticeable with high rank arrays. Performance data for ranks 2-6 would helpful in this case.

dholladay00:
The more I think about this the more I think we should recompute the strides. Even so, I'd like to see performance differences. Integer operations can be noticeable on GPUs so pre-computed may be more performant, but a larger object may use more registers. There are a lot of competing factors so data is the best.

dholladay00:
now that singularity-eos 1.9.0 release has been cut, I think we should up to c++17 and get this merged in prep for the next release.

Yurlungur:
> now that singularity-eos 1.9.0 release has been cut, I think we should up to c++17 and get this merged in prep for the next release.

I am in favor of moving to C++17 at this time. However, I would prefer to decouple the shift to 17 from this MR. Let's just update the cmake builds to require 17 and then merge other MRs when they are ready.

cmauney:
@jonahm-LANL @dholladay00 I've pushed some benchmarks using `catch2`, right now just doing index computation through contiguous memory - `(i, j, k)` to `n`. I haven't tested this on device yet but I would be interested to see the results. I intend to put in the same benchmarks on random (rather than contiguous) memory.

I can take off `Draft` and move forward with review if you want to get this in. There are things I'd like to modify and add but there's a danger of code-creep and I don't want to hold up, the PR 'as-is' should be ready to go.

ports-of-call/portable_arrays.hpp
```diff
@@ -26,73 +26,93 @@
...
+// `arr_mul({vars...})`
+template <typename T, std::size_t N>
+PORTABLE_INLINE_FUNCTION auto arr_mul(const std::array<T, N> &a) {
+  auto r = T{1};
+  for (auto v : a)
+    r *= v;
+  return r;
+}
+PORTABLE_FORCEINLINE_FUNCTION
+decltype(auto) vp_prod() {
```
Yurlungur, line 58: is the `decltype` needed? Why not just an `auto` return value?

ports-of-call/portable_arrays.hpp
```diff
@@ -262,20 +208,78 @@ class PortableMDArray {
                             const int indx, const int nvar);
 
  private:
+  template <typename... NX, std::size_t N = sizeof...(NX)>
+  PORTABLE_FORCEINLINE_FUNCTION auto make_nxs_array(NX... nxs) {
+    std::array<std::size_t, MAXDIM> a;
+    std::array<std::size_t, N> t{static_cast<std::size_t>(nxs)...};
+    for (auto i = 0; i < N; ++i) {
```
Yurlungur, line 215: I'm gonna insist this be `int` or `size_t`. IMO this is not a style thing but a self-documenting thing. It's an index it's not a double.

Yurlungur (commented):
I like this---it's a significant cleanup from the original implementation, which was quite low-level. I want @dholladay00 's build system concerns addressed before merge. Also, just to confirm, this doesn't change the API forfunctionality at all right? It's just more general.

cmauney:
If C++17 gets in before this is merged, then I do want to do another draft - it will make things cleaner and clearer.

ports-of-call/portable_arrays.hpp
mauneyc-LANL: eh, I'm returning a generic lambda and that's my habbit. But it's not necessary here (tho I don't think it's harmful either)

Yurlungur: I don't think the `decltype` adds anything and it's a bit harder to read but I won't push back too much on this.

Yurlungur:
@mauneyc-LANL let us know when this is ready for re-review

Yurlungur:
@mauneyc-LANL what's the status of this?

mauneyc-LANL:
@jonahm-LANL Code-proper is good but I'm fleshing out the test, benchmarks and making sure they work on other machines. That could be separated into another PR.

Yurlungur:
Sounds good. I think 1 test on github is enough to merge this, but I'd like at least one. We can do the more broad test set later down the line.

test/test_utilities.hpp
Yurlungur, line 1: :+1: these are nice

ports-of-call/utility/index_algo.hpp
```diff
@@ -0,0 +1,96 @@
...
+// Administration. The Government is granted for itself and others acting
+// on its behalf a nonexclusive, paid-up, irrevocable worldwide license
+// in this material to reproduce, prepare derivative works, distribute
+// copies to the public, perform publicly and display publicly, and to
+// permit others to do so.
+
+#ifndef _PORTSOFCALL_UTILITY_INDEX_ALGO_HPP_
+#define _PORTSOFCALL_UTILITY_INDEX_ALGO_HPP_
+
+#include "../portability.hpp"
```
Yurlungur, line 17: might be best to use `<ports-of-call/portability.hpp>`

ports-of-call/portable_arrays.hpp
```diff
@@ -24,438 +24,275 @@
...
+#endif // __cplusplus
+
+template <auto N>
+using IArray = Array<std::size_t, N>;
+
+template <typename T, std::size_t D = MAXDIM>
 class PortableMDArray {
  public:
-  static constexpr int MAXDIM = 6;
+  using this_type = PortableMDArray<T, D>;
```
Yurlungur, line 67: We should have some introspection on the maximum dimension of `PortableMDArray`. Can we add a member field that reports `D`?

ports-of-call/portable_arrays.hpp
```diff
@@ -24,438 +24,275 @@
...
-    nx1_ = nx1;
-    nx2_ = nx2;
-    nx3_ = nx3;
-    nx4_ = nx4;
-    nx5_ = nx5;
-    nx6_ = nx6;
+  PORTABLE_INLINE_FUNCTION size_type GetRank() const noexcept { return rank_; }
+  template <typename... NXs>
+  PORTABLE_INLINE_FUNCTION void Reshape(NXs... nxs) {
+    assert(util::array_reduce(IArray<D>{nxs...}, 1, std::multiplies<size_type>{}) ==
```
Yurlungur, line 187: I know this was an assert previously, but now we have `PORTABLE_REQUIRE`. Let's use that instead as it will provide more useful/verbose output.

Yurlungur (changes requested):
Some minor changes below. Also one high-level question: Given that this infrastructure assumes continuous data and only supports contiguous slices, is there a reason to include both sizes and strides? I think we only need to carry around sizes.

As discussed, we can skip benchmarking tests for now. Though I am somewhat concerned about performance.

Yurlungur:
Ah regarding strides, never mind. I see @dholladay00 's point above. I forgot about it.

Yurlungur:
@mauneyc-LANL I tried updating spiner to use your branch and spiner fails with a huge swath of test failures, including several segfaults. We need to resolve that before this branch can be merged. Branch `jmm/test-md-refactor` in spiner.

Yurlungur:
I think the issue is `fast_index` the original `PortableMDArray` implementation allows one to "pun" the dimension of the array to smaller sizes. In particular, indexing in 1d always accesses the flat index, which is a useful feature.

Yurlungur:
OK I fixed the segfault but several tests still failing. @mauneyc-LANL these need to be fixed before merge. That said, the SAP integration effort is a higher priority. If you're ok with it let's punt on this for now.

ports-of-call/portable_arrays.hpp
```diff
@@ -187,12 +187,28 @@ class PortableMDArray {
...
   template <typename... Is> //, class = std::enable_if_t<(sizeof...(Is) > 3)>>
   PORTABLE_FORCEINLINE_FUNCTION constexpr reference operator()(Is... idxs) noexcept {
-    return pdata_[util::fast_findex({static_cast<size_type>(idxs)...}, nxs_, strides_)];
+    index_type idx;
+    if constexpr (sizeof...(Is) == 0) {
+      idx = 0;
+    } else if constexpr (sizeof...(Is) == 1) {
+      idx = get_first(idxs...);
+    } else {
+      idx = util::fast_findex({static_cast<size_type>(idxs)...}, nxs_, strides_);
```
mauneyc-LANL: `fast_findex` already checks the size of the input indices. It's missing the empty case but it should be added there.

What is the difference between

```c++
fast_findex(A const &ijk, A const &dim, A const &stride) {
  constexpr auto N = get_size(A{});
  if constexpr (N == 1) {
    return ijk[0];
  }
 //...
}
``` 
and `get_first` ?

mauneyc-LANL (commented):
Can you clarify?

ports-of-call/portable_arrays.hpp
Yurlungur: You could instead modify fast index instead of adding get_first I think. Either should be fine.

mauneyc-LANL:
Need to rethink and reconfigure some of this work. Will try to split out the ideas into more focused PRs.

PR #40, opened 2024-01-23
First cut at adding printf, help builds.

dholladay00:
## PR Summary

This adds an internal print function used by portable errors machinery that stubs out printf on hip where printf doesn't exist.

- [ ] Should we make this more public for use by downstream codes (probably).
- [ ] Should we figure out a way to call out to kokkos printf if its available (only available on very new versions of kokkos).

@Yurlungur @mauneyc-LANL @rbberger @chadmeyer 

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-call/portability.hpp
+++ b/ports-of-call/portability.hpp
@@ -101,6 +101,15 @@ constexpr bool EXECUTION_IS_HOST{false};
 #else
 constexpr bool EXECUTION_IS_HOST{true};
 #endif
+// portable printf
+template <typename ... Ts>
+PORTABLE_INLINE_FUNCTION void printf(char const * const format, Ts ... ts) {
+  // disable for hip
+#ifndef __HIPCC__
+  std::printf(format, ts...);
+#endif // __HIPCC__
+  return;
+}
 } // PortsOfCall
 
 template <typename T>
--- a/ports-of-call/portable_errors.hpp
+++ b/ports-of-call/portable_errors.hpp
@@ -28,11 +28,6 @@
 // TODO(JMM): Is relative path right here?
 #include "portability.hpp"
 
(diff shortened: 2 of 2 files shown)


Yurlungur (commented):
This is a good add. Doesn't look like it's all working yet?

dholladay00:
This works sufficiently for our purposes as this change-set gets `help` to build on rzvernal @Yurlungur, the question is do we want to do a bunch of complicated stuff to add the kokkos version for sufficiently new versions of kokkos? If not (I am leaning in this direction) we can remove some comments. 

The other question @Yurlungur is should be remove this from impl namespace and put it in a public facing namespace? I think yes as downstream codes will likely want this thing if they want to build with hip.

Yurlungur:
Broadly, I concur:

> This works sufficiently for our purposes as this change-set gets `help` to build on rzvernal @Yurlungur, the question is do we want to do a bunch of complicated stuff to add the kokkos version for sufficiently new versions of kokkos? If not (I am leaning in this direction) we can remove some comments.

No---let's wait until that's just the version we depend on. We can do what you did here until then,
 
> The other question @Yurlungur is should be remove this from impl namespace and put it in a public facing namespace? I think yes as downstream codes will likely want this thing if they want to build with hip.

Yes, let's put it in the public namespace.

Yurlungur:
@dholladay00 let me know when this is ready for review/merge

dholladay00:
@Yurlungur are there additional tests on Darwin that I can run or is this sufficient? This is ready to go pending necessary testing if not already accomplished.

Yurlungur (approved):
I think we're good to go

PR #44, opened 2024-08-14
Change 'condition' to 'condition_bool'

### LINKED ISSUE #43: Bug in `require`
It looks like this line should contain `condition_bool` rather than `condition` twice:
https://github.com/lanl/ports-of-call/blob/c01fa56163098aacaa0eb02d751d2780d7dcef49/ports-of-call/portable_errors.hpp#L131

This affects the string stream implementation of `PORTABLE_REQUIRE`.

jhp-lanl:
## PR Summary

Quick fix to what appears to be a bug where `condition` was twice in the wrapper for `require()`. This affects the string stream implementation of `PORTABLE_REQUIRE`.

Fixes #43 

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.
- N/A If preparing for a new release, update the version in cmake.

### DIFF
--- a/ports-of-call/portable_errors.hpp
+++ b/ports-of-call/portable_errors.hpp
@@ -128,7 +128,7 @@ inline void require(bool condition_bool, const char *const condition,
 inline void require(bool condition_bool, const char *const condition,
                     std::stringstream const &message,
                     const char *const filename, int const linenumber) {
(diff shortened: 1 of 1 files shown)


Yurlungur (approved):
Oops good catch!

PR #46, opened 2024-08-26
Transfer static vector from Singe

BrendanKKrueger:
## PR Summary

A GPU-capable implementation of static vector.  This is a `std::vector`-like data structure built on top of a `std::array`-like backing store, resulting in something with a compile-time capacity but a run-time size.  It works on both the CPU and the GPU, and can be transferred between the two.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-call/static_vector.hpp
+++ b/ports-of-call/static_vector.hpp
@@ -318,11 +318,11 @@ class static_vector {
 
   // begin
   // --------------------------------------------------------------------------------------
-  PORTABLE_FUNCTION constexpr iterator begin() { return {ptr()}; }
+  PORTABLE_FUNCTION constexpr iterator begin() { return ptr(); }
 
-  PORTABLE_FUNCTION constexpr const_iterator begin() const { return {ptr()}; }
+  PORTABLE_FUNCTION constexpr const_iterator begin() const { return ptr(); }
 
-  PORTABLE_FUNCTION constexpr const_iterator cbegin() const { return {ptr()}; }
+  PORTABLE_FUNCTION constexpr const_iterator cbegin() const { return ptr(); }
 
   // end
   // ----------------------------------------------------------------------------------------
--- a/ports-of-call/static_vector.hpp
+++ b/ports-of-call/static_vector.hpp
@@ -359,18 +359,6 @@ class static_vector {
     count_ = 0;
   }
 
-  // insert
-  // -------------------------------------------------------------------------------------
-  // TODO: In order to simplify, I removed all versions of insert for now.
-
-  // emplace
-  // ------------------------------------------------------------------------------------
-  // TODO: In order to simplify, I removed all versions of emplace for now.
-
-  // erase
-  // --------------------------------------------------------------------------------------
-  // TODO: In order to simplify, I removed all versions of erase for now.
-
   // push_back
   // ----------------------------------------------------------------------------------
   PORTABLE_FUNCTION constexpr void push_back(value_type const &value) {
--- a/ports-of-call/static_vector.hpp
+++ b/ports-of-call/static_vector.hpp
@@ -1,0 +1,436 @@
+#ifndef _PORTS_OF_CALL_STATIC_VECTOR_HPP_
+#define _PORTS_OF_CALL_STATIC_VECTOR_HPP_
+
+#include "ports-of-call/portability.hpp"
+
+#include <cassert>
+#include <initializer_list>
+#include <iterator>
+#include <type_traits>
+
+namespace PortsOfCall {
+
+/* The static_vector has an interface modeled after std::vector, but it uses
+ * statically-allocated memory to (a) avoid dynamic memory allocation, deallocation, or
+ * reallocation; and (b) ensure that the data is contained entirely inside the
+ * static_vector so that this type can be memcopied to a CPU.  We had an initial
+ * implementation that provided additional features, but was not portable to GPUs.  This
+ * implementation is simpler and less feature-rich, but works on GPUs.  Tests have been
+ * added to the test suite to ensure that this is true and to enforce that it continues to
+ * be true.  Features that are missing:
+ * -- The insert, emplace, and erase methods from std::vector were included in the initial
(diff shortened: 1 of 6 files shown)


ports-of-call/static_vector.hpp
```diff
@@ -0,0 +1,436 @@
...
+      return other < *this;
+    }
+
+    PORTABLE_FUNCTION constexpr bool operator>=(const_iterator_type other) const {
+      return not(*this < other);
+    }
+
+    PORTABLE_FUNCTION constexpr bool operator<=(const_iterator_type other) const {
+      return not(*this > other);
+    }
```
Yurlungur, line 148: Are all of the comparators really necessary? Won't the compiler fill in the definitions of some of these if we don't specify?

ports-of-call/static_vector.hpp
```diff
@@ -0,0 +1,436 @@
...
+  // -------------------------------------------------------------------------------------
+  // TODO: In order to simplify, I removed all versions of insert for now.
+
+  // emplace
+  // ------------------------------------------------------------------------------------
+  // TODO: In order to simplify, I removed all versions of emplace for now.
+
+  // erase
+  // --------------------------------------------------------------------------------------
+  // TODO: In order to simplify, I removed all versions of erase for now.
```
Yurlungur, line 372: I'm not convinced these are needed ever. There's no reason we have to be 100% "standards compliant."

test/test_static_vector_kokkos.cpp
```diff
@@ -0,0 +1,113 @@
+#include "ports-of-call/portability.hpp"
+#include "ports-of-call/static_vector.hpp"
+
+#ifndef CATCH_CONFIG_FAST_COMPILE
```
Yurlungur, line 4: I know it's only included if you built with kokkos, but IMO it would still be good to wrap this in `ifdef`s just in case.

Yurlungur (approved):
This is pretty neat! Could you also add to the docs showing this capability exists?

ports-of-call/static_vector.hpp
BrendanKKrueger: I think you're thinking of the [spaceship operator](https://en.cppreference.com/w/cpp/language/default_comparisons), which a C++20 feature.

test/test_static_vector_kokkos.cpp
BrendanKKrueger: I'm not sure what `ifdef`(s) you're suggesting, sorry

dholladay00: I assume @Yurlungur is referring to the `PORTABILITY_STRATEGY_[KOKKOS|NONE|etc.]` macros.

ports-of-call/static_vector.hpp
```diff
@@ -0,0 +1,424 @@
...
+
+  // data
+  // ---------------------------------------------------------------------------------------
+  PORTABLE_FUNCTION constexpr pointer data() { return ptr(); }
+
+  PORTABLE_FUNCTION constexpr const_pointer data() const { return ptr(); }
+
+  // begin
+  // --------------------------------------------------------------------------------------
+  PORTABLE_FUNCTION constexpr iterator begin() { return {ptr()}; }
```
dholladay00, line 321: what is the purpose of the braces @BrendanKKrueger ?

ports-of-call/static_vector.hpp
```diff
@@ -0,0 +1,424 @@
...
+    assert(count_ < N);
+    new (ptr(count_)) value_type(std::forward<Args>(args)...);
+    ++count_;
+    return back();
+  }
+
+  // pop_back
+  // -----------------------------------------------------------------------------------
+  PORTABLE_FUNCTION constexpr void pop_back() {
+    if (count_ > 0) {
```
dholladay00, line 389: should this just be an assertion?

test/test_static_vector_kokkos.cpp
BrendanKKrueger: Oh, do you mean to just wrap the whole file in `#ifdef PORTABILITY_STRATEGY_KOKKOS`?

ports-of-call/static_vector.hpp
BrendanKKrueger: Probably cut-and-paste from an earlier implementation of static vector that someone else wrote.  Presumably it was intended to indicate that the pointer is an argument to the constructor of the iterator?  But it doesn't appear to be necessary, as I deleted the braces on line 321 and the code seemed to build and run fine.  Would you like me to strip all those extra braces?

BrendanKKrueger: PS: I say "that someone else wrote" because `git blame` in Singe claims I wrote the file.  I don't recognize a lot of this code, so either my memory sucks (always possible with me) or `git blame` is confused because I did a reformat or similar.  I can't answer for every decision in the code, but I'm happy to iterate with you to adapt the code to something that "fits" with Ports-of-Call.

dholladay00: My preference is to remove the braces if unneeded, but I can be convinced otherwise if there are good arguments for it.

ports-of-call/static_vector.hpp
dholladay00: Ah that makes sense.

ports-of-call/static_vector.hpp
BrendanKKrueger: Deleting the braces seemed to have no effect, so I'm not aware of any arguments to keep them.

Yurlungur:
@dholladay00 are you happy with this? If so will merge.

test/test_static_vector_kokkos.cpp
Yurlungur: yeah that's right

dholladay00:
> @dholladay00 are you happy with this? If so will merge.

yep 👍

dholladay00 (approved):

PR #47, opened 2024-08-26
Transfer span from Singe

BrendanKKrueger:
## PR Summary

An implementation of `std::span` that works on both the CPU and the GPU.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-call/span.hpp
+++ b/ports-of-call/span.hpp
@@ -12,10 +12,10 @@ namespace PortsOfCall {
 // Heavily simplified relative to std::span, but provides a similar concept in a way
 // that's portable to GPUs.
 
-template <typename DataType>
+template <typename T>
 class span {
  private:
-  DataType *ptr_{nullptr};
+  T *ptr_{nullptr};
   std::size_t size_{0};
 
  public:
@@ -33,7 +33,7 @@ class span {
   // object provides access to indices 2 <= i < 8 of the array my_array, but the indices
   // are shifted so that sub_range[i] is my_array[i+2].
   template <typename SizeType>
-  PORTABLE_FUNCTION constexpr span(DataType *ptr, SizeType const &count)
+  PORTABLE_FUNCTION constexpr span(T *ptr, SizeType const &count)
       : ptr_{ptr}, size_(count) {
     assert(count >= 0);
   }
@@ -42,26 +42,26 @@ class span {
   PORTABLE_FUNCTION constexpr auto size() const { return size_; }
 
   // Iterator (really a pointer) to the beginning of the range, providing mutable access.
-  PORTABLE_FUNCTION constexpr DataType *begin() { return ptr_; }
+  PORTABLE_FUNCTION constexpr T *begin() { return ptr_; }
 
   // Iterator (really a pointer) to the beginning of the range, providing constant access.
-  PORTABLE_FUNCTION constexpr DataType const *begin() const { return ptr_; }
+  PORTABLE_FUNCTION constexpr T const *begin() const { return ptr_; }
 
   // Iterator (really a pointer) to the beginning of the range, providing constant access.
-  PORTABLE_FUNCTION constexpr DataType const *cbegin() const { return ptr_; }
+  PORTABLE_FUNCTION constexpr T const *cbegin() const { return ptr_; }
 
   // Iterator (really a pointer) to the end of the range, providing mutable access.
-  PORTABLE_FUNCTION constexpr DataType *end() { return ptr_ + size_; }
+  PORTABLE_FUNCTION constexpr T *end() { return ptr_ + size_; }
 
   // Iterator (really a pointer) to the beginning of the range, providing constant access.
-  PORTABLE_FUNCTION constexpr DataType const *end() const { return ptr_ + size_; }
+  PORTABLE_FUNCTION constexpr T const *end() const { return ptr_ + size_; }
 
   // Iterator (really a pointer) to the beginning of the range, providing constant access.
-  PORTABLE_FUNCTION constexpr DataType const *cend() const { return ptr_ + size_; }
+  PORTABLE_FUNCTION constexpr T const *cend() const { return ptr_ + size_; }
 
   // Index operator to obtain mutable access to an element of the range.
   template <typename Index>
-  PORTABLE_FUNCTION constexpr DataType &operator[](Index const &index) {
+  PORTABLE_FUNCTION constexpr T &operator[](Index const &index) {
     assert(index >= 0);
     assert(static_cast<std::size_t>(index) < size_);
     return *(ptr_ + index);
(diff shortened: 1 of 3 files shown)


ports-of-call/span.hpp
```diff
@@ -0,0 +1,91 @@
...
+// Heavily simplified relative to std::span, but provides a similar concept in a way
+// that's portable to GPUs.
+
+template <typename DataType>
+class span {
+ private:
+  DataType *ptr_{nullptr};
+  std::size_t size_{0};
+
+ public:
```
mauneyc-LANL, line 21: As a corollary to the above, having some traits with verbose naming is also beneficial. As a suggestion:

```c++
using element_type = T;
using value_type = std::remove_cv_t<T>;
using index_type = std::ptrdiff_t; 
using difference_type = index_type;
using pointer = T*;
using reference = T&;
using iterator = T*;
using reverse_iterator =
    std::reverse_iterator<iterator>;
```

ports-of-call/span.hpp
```diff
@@ -0,0 +1,91 @@
...
+  // Construct an empty span
+  PORTABLE_FUNCTION constexpr span() = default;
```
mauneyc-LANL, line 23: `PORTABLE_FUNCTION constexpr span() noexcept = default` and for other functions also

ports-of-call/span.hpp
```diff
@@ -0,0 +1,91 @@
...
+  // For example:
+  //      double* my_array = new double[10]
+  //      span full_range(my_array, 10); // elements 0-9 (all 10 elements in the array)
+  //      span sub_range(my_array+2, 6); // elements 2-7 (total of six elements)
+  // The full_range object provides access to the full range of my_array.  The sub_range
+  // object provides access to indices 2 <= i < 8 of the array my_array, but the indices
+  // are shifted so that sub_range[i] is my_array[i+2].
+  template <typename SizeType>
+  PORTABLE_FUNCTION constexpr span(DataType *ptr, SizeType const &count)
+      : ptr_{ptr}, size_(count) {
```
mauneyc-LANL, line 37: Pass `count` by value. Generally, integer types are no larger than pointers.

mauneyc-LANL (changes requested):
I have a few changes and/or points of discussion.

In general, I'm okay with this as a placeholder awaiting the move to C++20. If it's of interest, I have [some code](https://github.com/cmauney/nanda/blob/main/libs/nanda/include/nanda/span.hh) with a more "compliant" `span` (the code is unfinished, but `span` is finished).

There is an general question with the introduction of `span`; the next obvious step is `mdspan`, but this would basically supersede `PortableMDArray` (and, in my opinion, this should happen). Is this the goal?

ports-of-call/span.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <cassert>
+#include <cstddef>
+
+namespace PortsOfCall {
+
+// ================================================================================================
+// Heavily simplified relative to std::span, but provides a similar concept in a way
+// that's portable to GPUs.
+
+template <typename DataType>
```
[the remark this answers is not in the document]
BrendanKKrueger: Singe has enough templates that it makes sense to be more verbose.  In this case, there's only a single template and it's the data type.  People are used to seeing `template<typename T>` for containers like this, so it's not likely to confuse anyone.  I can make this change.

ports-of-call/span.hpp
BrendanKKrueger: These look like the standard typedefs for C++ containers, so that makes sense.

BrendanKKrueger:
> I have a few changes and/or points of discussion.
> 
> In general, I'm okay with this as a placeholder awaiting the move to C++20. If it's of interest, I have [some code](https://github.com/cmauney/nanda/blob/main/libs/nanda/include/nanda/span.hh) with a more "compliant" `span` (the code is unfinished, but `span` is finished).
> 
> There is an general question with the introduction of `span`; the next obvious step is `mdspan`, but this would basically supersede `PortableMDArray` (and, in my opinion, this should happen). Is this the goal?

* `span`: This version of `span` (actually called `span_like` in Singe) was specifically tailored to only implement the bits of `span` that Singe needed.  If you have a more complete / more compliant version of `span`, that would probably be better than this.  I wasn't putting this into Ports-of-Call because it's a high-quality, GPU-compatible implementation of `span`, but because it's something Singe has that other XCAP libraries may find useful.  And if something better than this already exists then I have zero concerns about throwing this out and using the better version so long as it covers all the necessary use-cases.
* `mdspan`: In principle I like the idea of aiming for something that will eventually be replaced by a standard container, but I've not looked at `PortableMDArray` so I can't comment on any differences between them.

ports-of-call/span.hpp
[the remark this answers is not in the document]
Yurlungur: :+1: I prefer `T` here.

ports-of-call/span.hpp
Yurlungur: I believe `PORTABLE_FUNCTION` is *not* required for `defaulted` functions. In fact including the annotation produces warnings.

Yurlungur:
> In general, I'm okay with this as a placeholder awaiting the move to C++20. If it's of interest, I have [some code](https://github.com/cmauney/nanda/blob/main/libs/nanda/include/nanda/span.hh) with a more "compliant" `span` (the code is unfinished, but `span` is finished).

I would be supportive of a portable backport of `span` into `ports-of-call` if this is useful for singe. The point of @BrendanKKrueger 's contribution was as he says, just that it's minimal---only what we needed.

> There is an general question with the introduction of `span`; the next obvious step is `mdspan`, but this would basically supersede `PortableMDArray` (and, in my opinion, this should happen). Is this the goal?

This is not the current goal. I'm open to this path, but for backwards compatibility reasons would prefer `PortableMDArray` wrap `MDSpan` if we go that path.

I *think* the only things that currently use `PortableMDArray` are `spiner` and tests` so they could be migrated. But I'd prefer to do it piece by piece. 

All that said: our priority is not cleaning up the old code in `ports-of-call`. It's not perfect but it's good enough. Our priority is making sure we are supporting the needs of the libraries that depend on us in a performant way that does not compromise code maintainability.

In the long term, I'm supportive of replacing `PortableMDarray` with a performance portable `mdspan` backport. If providing performance portable`mdspan` will save code duplication accross the XCAP libraries, then that's something we can prioritize. But if not then it needs to be on the backburner vs more urgent priorities.

ports-of-call/span.hpp
BrendanKKrueger: I used the standard typedefs from `std::span` (https://en.cppreference.com/w/cpp/container/span), so it's slightly different from what you suggested.  I checked, and `std::reverse_iterator` is `constexpr` as of C++17, so that works.

ports-of-call/span.hpp
BrendanKKrueger: I made both suggested changes.

Yurlungur: I'm confused---aren't these suggestions contradictory?

BrendanKKrueger: I read the comments as:
* Add `noexcept`
* Remove `PORTABLE_FUNCTION`

Yurlungur (approved):
LGTM at this point. @mauneyc-LANL please re-review.

mauneyc-LANL:
I'm closing this as it is superseded by https://github.com/lanl/ports-of-call/pull/60.

PR #48, opened 2024-08-26
Transfer array from Singe

BrendanKKrueger:
## PR Summary

An implementation of `std::array` that works on both the CPU and the GPU.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-call/array.hpp
+++ b/ports-of-call/array.hpp
@@ -1,0 +1,270 @@
+#ifndef _PORTS_OF_CALL_ARRAY_HPP_
+#define _PORTS_OF_CALL_ARRAY_HPP_
+
+#include "portability.hpp"
+#include "portable_errors.hpp"
+
+#include <cstddef>
+#include <iterator>
+#include <tuple>
+
+namespace PortsOfCall {
+
+//! The array class is a constexpr (mostly) replacement for std::array
+/*!
+std::array is not fully constexpr until C++20.  Currently we are relying on
+"constexpr"-ness to offload to GPUs.  PortsOfCall::array provides a
+constexpr replacement for std::array with the following notable exceptions:
+    - the reverse iterators are not provided because std::reverse_iterator is
+not constexpr until C++17 (could be provided with modest effort)
+
+    - the `at` function is not provided because it bounds checks and throws an
+exception which we cannot do on a GPU (it could be added and made
+non-constexpr)
+
+    - no std::get (cannot be provided without undefined behavior)
+*/
+
+namespace array_detail {
+
+template <typename T, std::size_t N>
+struct bits {
+  using type = T[N];
+
+  template <typename IntT>
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T &ref_access(type &array, IntT index) {
+    return array[index];
+  }
+
+  template <typename IntT>
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T const &ref_access(type const &array,
+                                                                     IntT index) {
+    return array[index];
+  }
+
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T *ptr_access(type &array) {
+    return static_cast<T *>(array);
+  }
+
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T const *ptr_access(type const &array) {
+    return static_cast<T const *>(array);
+  }
+};
+
+template <typename T>
+struct bits<T, 0> {
+  using type = T[1];
+
+  template <typename IntT>
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T &ref_access(type &arr, IntT) {
+    return arr[0];
+  }
+
+  template <typename IntT>
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T const &ref_access(type const &arr,
+                                                                     IntT) {
+    return arr[0];
+  }
+
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T *ptr_access(type &arr) {
+    return (T *)&arr;
+  }
+
+  PORTABLE_FORCEINLINE_FUNCTION static constexpr T const *ptr_access(type const &arr) {
+    return (T const *)&arr;
+  }
+};
+} // namespace array_detail
+
+template <typename T, std::size_t N>
+struct array {
+  // member types
+  using value_type = T;
+  using size_type = std::size_t;
+  using difference_type = std::ptrdiff_t;
+  using reference = T &;
+  using const_reference = T const &;
+  using pointer = value_type *;
+  using const_pointer = value_type const *;
+  using iterator = pointer;
+  using const_iterator = const_pointer;
+
+  //! The  array being wrapped by this class
+  typename array_detail::bits<T, N>::type arr{};
+
+  //! Index operator that returns a reference to the desired element.
+  PORTABLE_FUNCTION constexpr reference operator[](size_type index) {
+    return array_detail::bits<T, N>::ref_access(arr, index);
+  }
+
+  //! bounds-checked equivalent to index operator
+  PORTABLE_FUNCTION constexpr reference at(size_type index) {
+    if (index >= size()) {
+      PORTABLE_ALWAYS_THROW_OR_ABORT("invalid index.");
+    }
+    return array_detail::bits<T, N>::ref_access(arr, index);
+  }
+
+  //! Index operator that returns a const reference to the desired element.
+  PORTABLE_FUNCTION constexpr const_reference operator[](size_type index) const {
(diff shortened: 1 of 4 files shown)


mauneyc-LANL (approved):

ports-of-call/array.hpp
```diff
@@ -0,0 +1,262 @@
...
+/*!
+std::array is not fully constexpr until C++20.  Currently we are relying on
+"constexpr"-ness to offload to GPUs.  PortsOfCall::array provides a
+constexpr replacement for std::array with the following notable exceptions:
+    - the reverse iterators are not provided because std::reverse_iterator is
+not constexpr until C++17 (could be provided with modest effort)
+
+    - the `at` function is not provided because it bounds checks and throws an
+exception which we cannot do on a GPU (it could be added and made
+non-constexpr)
```
Yurlungur, line 22: I would suggest providing `at` with `PORTABLE_ALWAYS_THROW_OR_ABORT` which will throw an exception on host, but not device based on the protability strategy. (Requires including `portability.hpp` and `portable_errors.hpp`)

ports-of-call/array.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+    - no std::get (cannot be provided without undefined behavior)
+*/
+
+namespace array_detail {
+
+template <typename T, std::size_t N>
+struct bits {
```
Yurlungur, line 30: I must be missing something---I don't understand the point of this container struct.

ports-of-call/array.hpp
```diff
@@ -0,0 +1,262 @@
...
+  using type = T[N];
```
Yurlungur, line 31: Would prefer this be `Array_t` rather than `type`

ports-of-call/array.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+  template <typename IntT>
+  PORTABLE_FUNCTION static constexpr T &ref_access(type &array,
```
Yurlungur, line 34: @dholladay00 should correct me if I'm wrong. Do we need `PORTABLE_FUNCTION` decorators for `constexpr`? Either way I think `PORTABLE_FORCEINLINE_FUNCTION` is better here than `PORTABLE_FUNCTION`.

test/test_array.hpp
```diff
@@ -0,0 +1,235 @@
...
+
+  // Fill the array
+  for (int i = 0; i < N; ++i) {
+    arr[i] = i + 1;
+  }
+
+  // Can we read from it?
+  for (int i = 0; i < N; ++i) {
+    CHECK(arr[i] == i + 1);
+  }
```
Yurlungur, line 20: Some of these loops should be inside device kernels (e.g., `portable_for`) to ensure it works on device.

test/test_portsofcall.cpp
```diff
@@ -18,6 +18,8 @@
 #define CATCH_CONFIG_RUNNER
 #include "catch2/catch.hpp"
 
+#include "test_array.hpp"
```
Yurlungur, line 21: Not a fan of this... We should probably just copy the setup we have in singularity to do multiple test files in individual `cpp` files.

Yurlungur (changes requested):
Big picture I think this is a great addition and would like to see this included. Given discussion with @mauneyc-LANL and @rbberger I think we cna merge without CI actually running, so long as you ran tests locally and we can wait for the great test rewrite from @rbberger .

A few comments/questions/requested changes below. Feel free to push back on anything.

Yurlungur:
Oh also there should be at least a mention this exists in the docs.

ports-of-call/array.hpp
mauneyc-LANL: `Array_t` is somewhat quixotic; `type` is standard trait naming.

mauneyc-LANL:
I was digging around a bit and found this

https://github.com/NVIDIA/libcudacxx/issues/51#issuecomment-717984614

TLDR: 1) `--expt-relaxed-constexpr` & `-std=c++17` seems to provide sufficient functionality for `std::array` on device, 2.) `libcu++` rolls a `std::array` in `cuda/std/array` (as of 2023).

Would this make a hand-rolled `std::array` necessary?

Yurlungur:
> I was digging around a bit and found this
> 
> [NVIDIA/libcudacxx#51 (comment)](https://github.com/NVIDIA/libcudacxx/issues/51#issuecomment-717984614)
> 
> TLDR: 1) `--expt-relaxed-constexpr` & `-std=c++17` seems to provide sufficient functionality for `std::array` on device, 2.) `libcu++` rolls a `std::array` in `cuda/std/array` (as of 2023).
> 
> Would this make a hand-rolled `std::array` necessary?

:+1: This is what we do in parthenon. But is that fully portable?

ports-of-call/array.hpp
BrendanKKrueger: No idea -- this was CJ code.  I can only assume that this allows you to not have to handle special cases for `N = 0`?

ports-of-call/array.hpp
BrendanKKrueger: Several years ago we had some issues where `constexpr` wasn't completely sufficient.  It may have just been a quirk of an older compiler or something like that, and maybe they're redundant now.

Yurlungur: no I think the decorator may still be desirable... but since it's inline no reason not to use `PORTABLE_FORCEINLINE_FUNCTION` rather than `PORTABLE_FUNCTION`

BrendanKKrueger:
> > I was digging around a bit and found this
> > [NVIDIA/libcudacxx#51 (comment)](https://github.com/NVIDIA/libcudacxx/issues/51#issuecomment-717984614)
> > TLDR: 1) `--expt-relaxed-constexpr` & `-std=c++17` seems to provide sufficient functionality for `std::array` on device, 2.) `libcu++` rolls a `std::array` in `cuda/std/array` (as of 2023).
> > Would this make a hand-rolled `std::array` necessary?
> 
> 👍 This is what we do in parthenon. But is that fully portable?

_Most_ of `std::array` works on GPUs with relaxed `constexpr`, but not all.  In particular, `std::array::fill` and `std::array::swap` aren't `constexpr` until C++20 and I believe Singe code uses `std::array:fill`.  Once we move to C++20 (hopefully much faster than the 14-to-17 transition), then this code becomes redundant if we're comfortable relying on relaxed `constexpr` to get the STL onto the GPU.

Yurlungur:
> > > I was digging around a bit and found this
> > > [NVIDIA/libcudacxx#51 (comment)](https://github.com/NVIDIA/libcudacxx/issues/51#issuecomment-717984614)
> > > TLDR: 1) `--expt-relaxed-constexpr` & `-std=c++17` seems to provide sufficient functionality for `std::array` on device, 2.) `libcu++` rolls a `std::array` in `cuda/std/array` (as of 2023).
> > > Would this make a hand-rolled `std::array` necessary?
> > 
> > 
> > 👍 This is what we do in parthenon. But is that fully portable?
> 
> _Most_ of `std::array` works on GPUs with relaxed `constexpr`, but not all. In particular, `std::array::fill` and `std::array::swap` aren't `constexpr` until C++20 and I believe Singe code uses `std::array:fill`. Once we move to C++20 (hopefully much faster than the 14-to-17 transition), then this code becomes redundant if we're comfortable relying on relaxed `constexpr` to get the STL onto the GPU.

that's a satisfying explanation for me. :+1:

BrendanKKrueger:
This is ready for another review

test/test_array.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+TEST_CASE("array nominal element access (GPU)", "[array][GPU]") {
+  // Declare the array
+  constexpr int N = 16;
+  array<double, N> arr;
+
+  int count = 0;
+  auto func = PORTABLE_LAMBDA(const int i, int &count) mutable {
+    // Fill the array
+    arr[i] = i + 1;
```
Yurlungur, line 21: I think the fill has to be done on host but it can be checked on device, because lambda will do a const capture.

Yurlungur (approved):
Looks great! Only one concern in the tests otherwise I think this is good to go.

ports-of-call/array.hpp
```diff
@@ -0,0 +1,270 @@
...
+#define _PORTS_OF_CALL_ARRAY_HPP_
+
+#include "portability.hpp"
+#include "portable_errors.hpp"
+
+#include <cstddef>
+#include <iterator>
+#include <tuple>
+
+namespace PortsOfCall {
```
BrendanKKrueger, line 11: @Yurlungur, is this namespacing correct?

Yurlungur: Yep :+1:

PR #60, opened 2024-10-08
Conformant span

mauneyc-LANL:
## PR Summary

A `span` implementation that (in broad terms) follows C++20. See https://en.cppreference.com/w/cpp/container/span for more information on this.

This replaces the PR provided by @BrendanKKrueger (https://github.com/lanl/ports-of-call/pull/47) in giving a more complete `span`.

@jonahm-LANL @dholladay00 

TODO:
- [x] Implement Testing
- [x] Cleaner comments

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-call/span.hpp
+++ b/ports-of-call/span.hpp
@@ -217,31 +217,18 @@ class span {
             span_REQUIRES(
                 (E == dynamic_extent || N == E) &&
                 detail::is_compatible_container<element_type (&)[N], ElementType>::value)>
-  // typename std::enable_if<(E == dynamic_extent || N == E) &&
-  //                             detail::is_compatible_container<
-  //                                 element_type (&)[N], ElementType>::value,
-  //                         int>::type = 0>
   constexpr span(element_type (&arr)[N]) noexcept : storage_(arr, N) {}
 
   template <class T, std::size_t N, std::size_t E = Extent,
             span_REQUIRES(
                 (E == dynamic_extent || N == E) &&
                 detail::is_compatible_container<std::array<T, N> &, ElementType>::value)>
-
-  // typename std::enable_if<(E == dynamic_extent || N == E) &&
-  //                             detail::is_compatible_container<
-  //                                 std::array<T, N> &, ElementType>::value,
-  //                         int>::type = 0>
   constexpr span(std::array<T, N> &arr) noexcept : storage_(arr.data(), N) {}
 
   template <class T, std::size_t N, std::size_t E = Extent,
             span_REQUIRES((E == dynamic_extent || N == E) &&
                           detail::is_compatible_container<const std::array<T, N> &,
                                                           ElementType>::value)>
-  // typename std::enable_if<(E == dynamic_extent || N == E) &&
-  //                             detail::is_compatible_container<
-  //                                 const std::array<T, N> &, ElementType>::value,
-  //                         int>::type = 0>
   constexpr span(const std::array<T, N> &arr) noexcept : storage_(arr.data(), N) {}
 
   // constructs a span that is a view of cont
@@ -254,10 +241,6 @@ class span {
       class Container, std::size_t E = Extent,
       span_REQUIRES(E == dynamic_extent && detail::is_container<Container>::value &&
                     detail::is_compatible_container<Container &, ElementType>::value)>
-  // typename std::enable_if<
-  //     E == dynamic_extent && detail::is_container<Container>::value &&
-  //         detail::is_compatible_container<Container &, ElementType>::value,
-  //     int>::type = 0>
   constexpr span(Container &cont) noexcept
       : storage_(detail::data(cont), detail::size(cont)) {}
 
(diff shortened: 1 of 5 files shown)


mauneyc-LANL:
Basically done. Let me know if you want something more or changed. @Yurlungur @dholladay00 @BrendanKKrueger

ports-of-call/span.hpp
```diff
@@ -0,0 +1,464 @@
...
+#if defined(__cpp_lib_span) // do we already have std::span?
+
+#include <span>
+namespace PortsOfCall::span {
+
+using std::span;
+
+} // namespace PortsOfCall::span
+
+#else
```
Yurlungur, line 35: Is `std::span` all constexpr? Will it work on device?

ports-of-call/span.hpp
```diff
@@ -0,0 +1,464 @@
...
+  constexpr span_storage() noexcept = default;
+
+  constexpr span_storage(E *p_ptr, std::size_t p_size) noexcept
+      : ptr(p_ptr), size(p_size) {}
+
+  E *ptr = nullptr;
+  std::size_t size = 0;
+};
+
+// some metaprogramming
```
Yurlungur, line 81: Is this stuff all in support of concepts?

ports-of-call/span.hpp
```diff
@@ -0,0 +1,464 @@
...
+} // namespace detail
+
+template <class ElementType, std::size_t Extent>
+class span {
+  static_assert(std::is_object<ElementType>::value,
+                "A span's ElementType must be an object type (not a "
+                "reference type or void)");
+  static_assert(detail::is_complete<ElementType>::value,
+                "A span's ElementType must be a complete type (not a forward "
+                "declaration)");
```
Yurlungur, line 162: Why is this?

ports-of-call/span.hpp
```diff
@@ -0,0 +1,464 @@
...
+template <class T>
+struct is_complete<T, decltype(sizeof(T))> : std::true_type {};
+
+} // namespace detail
+
+template <class ElementType, std::size_t Extent>
+class span {
+  static_assert(std::is_object<ElementType>::value,
+                "A span's ElementType must be an object type (not a "
+                "reference type or void)");
```
Yurlungur, line 159: And why is this?

test/test_span.cpp
Yurlungur, line 1: We should have some checks that span runs on device if possible.

Yurlungur (changes requested):
Thanks @mauneyc-LANL this looks really good! Some minor questions with two moderate asks:
1. Can you add a comment in the docs that we have this backport?
2. Can you add some tests that show it compiles/runs on device with Kokkos?

ports-of-call/span.hpp
mauneyc-LANL: Yes. I had wanted have more verbose concepts, but it was taking too long. What is here is a minimal set of constraints that cover most cases.

ports-of-call/span.hpp
mauneyc-LANL: See https://en.cppreference.com/w/cpp/container/span

> T 	- element type; must be a complete object type that is not an abstract class type

See also:

```c++
// detect if T is a complete type
// NB: sizeof(T) cannot be used with incomplete types
template <class, class = std::size_t>
struct is_complete : std::false_type {};

template <class T>
struct is_complete<T, decltype(sizeof(T))> : std::true_type {};
```

ports-of-call/span.hpp
mauneyc-LANL: contractually, `span` provides `size_bytes() { return size() * sizeof(element_type)`. `sizeof(void)` is undefined, and `sizeof(T&)` is in general incorrect. Further, `span` defines `reference_type = element_type &`, and things can get confusing very quickly. 

Disallowing references is a deeper topic, but basically has to do with how associated iterator types are determined, see https://en.cppreference.com/w/cpp/iterator/iter_t for a fun time

ports-of-call/span.hpp
Yurlungur: We should consider at some point either backporting concepts more fully into ports of call, or just waiting for C++20

ports-of-call/span.hpp
Yurlungur: In other words, this is standards specified. Ok.

ports-of-call/span.hpp
mauneyc-LANL: Based on what I was doing, I'd say it's not much worth it to try and replicate `std::concepts` and the like (or stuff that exists in the standard).

That said, I think we should start prototyping concepts for `singularity-eos` and `spiner`. What these would be should be given some thought, but it would be good to start thinking about how we want to use language-supported `concepts` and `requires` when they become available.

Yurlungur: agreed

test/test_span.cpp
mauneyc-LANL: Committed portability test

Yurlungur:
Looks like last thing is just a sentence in the docs saying this exists. THen I'm happy to merge. :+1:

Yurlungur (approved):

Yurlungur:
Looks like there's a merge conflict. Other than that, looks good to me

PR #61, opened 2024-10-09
Add config_summary to Ports-of-Call.

BrendanKKrueger:
## PR Summary

A recent addition to Singe.  Jonah said Ports-of-Call might find it useful, so putting together a demo so y'all can see if you like it or not.

@Yurlungur @dholladay00 @cmauney 

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,3 +144,28 @@ export(
   FILE "${CMAKE_CURRENT_BINARY_DIR}/${POCLIB}Targets.cmake")
 
 
+# Configuration summary ---------------------------------------------------------------------------
+
+include(config_summary)
+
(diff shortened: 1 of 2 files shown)


BrendanKKrueger:
By the way, if you want to see how this looks for Singe as an example, Diego has a screenshot [here](https://re-git.lanl.gov/xcap/ec/singe/-/merge_requests/438#note_974126).

Yurlungur (approved):
THis is nice. I think @mauneyc-LANL at some point tried to implement something like this but maybe not in ports.

mauneyc-LANL (approved):

BrendanKKrueger:
Let me know if these are the right options to display for Ports-of-Call, because I wasn't sure.

Yurlungur:
It seems totally reasonable to me.

PR #64, opened 2024-11-14
Fix non-constexpr math function

### LINKED ISSUE #63: Non constexpr expressions in tests
Tests fail to compile due to non-`constexpr` parts:
```
[ 85%] Building CXX object test/CMakeFiles/test_portsofcall.dir/test_static_vector.cpp.o
oss/ports-of-call/test/test_span.cpp:567:24: error: constexpr variable 'pi' must be initialized by a constant expression
  567 |   constexpr const Real pi = std::acos(-1);
      |                        ^    ~~~~~~~~~~~~~
/opt/rh/gcc-toolset-12/root/usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/cmath:100:14: note: subexpression not valid in a constant expression
  100 |     { return __builtin_acos(__x); }
      |              ^
oss/ports-of-call/test/test_span.cpp:567:29: note: in call to 'acos(-1)'
  567 |   constexpr const Real pi = std::acos(-1);
oss/ports-of-call/test/test_span.cpp:569:18: error: constexpr variable 'd_gp' must be initialized by a constant expression
  569 |   constexpr Real d_gp = 2. * pi / static_cast<Real>(N);
      |                  ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
oss/ports-of-call/test/test_span.cpp:569:30: note: initializer of 'pi' is not a constant expression
  569 |   constexpr Real d_gp = 2. * pi / static_cast<Real>(N);
      |                              ^
oss/ports-of-call/test/test_span.cpp:567:24: note: declared here
  567 |   constexpr const Real pi = std::acos(-1);
      |                        ^
2 errors generated when compiling for gfx942.
```

mauneyc-LANL:
## PR Summary

To fix #63 

Replaced "in-line" pi calculation with explicit decimal value.

@rbberger @jonahm-LANL

### DIFF
--- a/test/test_span.cpp
+++ b/test/test_span.cpp
@@ -1,7 +1,7 @@
-#include "ports-of-call/span.hpp"
 #include "ports-of-call/portability.hpp"
-#include <type_traits>
+#include "ports-of-call/span.hpp"
 #include <cmath>
+#include <type_traits>
 
 // ========================================================================================
 // © (or copyright) 2019-2024. Triad National Security, LLC. All rights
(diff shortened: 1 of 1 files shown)


test/test_span.cpp
```diff
@@ -564,19 +564,17 @@ TEST_CASE("span portability", "[PortsOfCall::span]") {
   constexpr const std::size_t Nb = N * sizeof(Real);
   constexpr const Real tol = 1.0E-8;
 
-  constexpr const Real pi = std::acos(-1);
+  constexpr const Real pi = 3.141592653589793238462643383279502884;
```
Yurlungur, line 567: Why not just use `M_PI`?

Yurlungur (approved):

test/test_span.cpp
mauneyc-LANL: Paranoid now about portability in the libraries. If you prefer `M_PI` let me know.

Yurlungur: All things being equal, I prefer `M_PI`, which is a macro set to that decimal isn't it? But if you think there's reasons to be paranoid, let's talk.

PR #65, opened 2024-11-19
fix config_summary

BrendanKKrueger:
## PR Summary

Export the Catch2 and Kokkos version variables up a level so that the config_summary machinery can find them.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


Yurlungur (approved):
thanks for fixing this!

PR #66, opened 2024-11-20
Improve warnings

BrendanKKrueger:
## PR Summary

* Eliminate some existing warnings
  * Some variables in the tests are only used for compile-time checks, and thus trip over unused-variable warnings.  Decorate those with `[[maybe_unused]]`.
  * Eliminated warnings about calling `__host__`-only functions from `__host__ __device__` functions by replacing `std::vector` in some tests with a very simple non-trivial type.
  * Modify `array::swap` to be device-capable (`std::swap` is not `constexpr` until C++20).
  * Because the actual type used for `std::size_t` is platform and compiler dependent, @mauneyc-LANL would sometimes see warnings about comparing unsigned integers against zero.  Replaced the checks that I could find with a small function that won't trip the warnings.
* Copied the warning flags from Nautilus: adds more warnings, promotes warnings to errors.
* Miscellaneous small changes for the new warning flags.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


test/CMakeLists.txt
```diff
@@ -56,6 +56,41 @@ elseif (PORTS_OF_CALL_TEST_PORTABILITY_STRATEGY STREQUAL "Cuda")
...
+    # TODO: find comiler flags for non-GNU compatible compilers
+    target_compile_options(portsofcall_iface INTERFACE
+        $<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},GNU>:-Werror;-Wall;-pedantic-errors;-Wunused-parameter>
+        $<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},AppleClang>:-Werror;-Wall;-Wunused-parameter>
+        $<$<NOT:$<BOOL:${WIN32}>>:$<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},Clang>:-Werror;-Wall;-Wno-unused-command-line-argument>>
+        $<$<BOOL:${WIN32}>:$<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},Clang>:-Wall;-Wno-c++98-compat;-Wno-c++98-compat-pedantic>>
+        $<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},Intel>:-Werror;-w2;-Wunused-variable;-Wunused-parameter;-diag-disable=remark>
+        $<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},PGI>:>
+        $<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},MSVC>:/permissive->
+        $<$<IN_LIST:${CMAKE_CXX_COMPILER_ID},XL>:>
```
Yurlungur, line 76: awesome how every compiler does this differently lol

Yurlungur (changes requested):
Thanks for this cleanup! As soon as tests pass, I'm happy. I think the current tests issue is that `span_EXPECTS` needs to typedef'd to be defined explicitly to a no-op for non-debug builds to avoid this warning (now promoted to error):
```C++
#define span_EXPECTS(...) assert((__VA_ARGS__))
```

test/CMakeLists.txt
BrendanKKrueger: Truly thrilling.  There's a reason I just cut-and-paste from library to library, after having figured it out to a "good enough" state once

BrendanKKrueger: PS: Also, `-Wall` (or whatever notation is used by a given compiler) is _not_ standardized and is likely gives a different set of warnings on every compiler...

Yurlungur: lol of course.

Yurlungur (approved):

PR #71, opened 2024-11-22
Add snprintf. Hide sprintf as fancy snprintf

Yurlungur:
## PR Summary

Apparently compilers don't like us using `sprintf` because of the risk of writing arbitrary bits to memory. This switches `portable_errors` to using `snprintf` for error messages.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/ports-of-call/portability.hpp
+++ b/ports-of-call/portability.hpp
@@ -127,7 +127,7 @@ inline std::size_t strlen(Head h, Tail... tail) {
       }
     }
   } else {
-    i = 20; // some big number to account for things like %.14e
+    i = 100; // some big number to account for things like %.14e
   }
   if constexpr (sizeof...(Tail) > 0) {
     i += strlen(tail...);
--- a/ports-of-call/portability.hpp
+++ b/ports-of-call/portability.hpp
@@ -117,7 +117,7 @@ PORTABLE_INLINE_FUNCTION void printf(char const *const format, Ts... ts) {
 }
 // Variadic strlen
 template <typename Head, typename... Tail>
-inline std::size_t strlen(Head h, Tail... tail) {
+PORTABLE_INLINE_FUNCTION std::size_t strlen(Head h, Tail... tail) {
   constexpr std::size_t MAX_I = 4096;
   std::size_t i = 0;
   if constexpr (std::is_convertible_v<Head, std::string_view>) {
--- a/ports-of-call/portability.hpp
+++ b/ports-of-call/portability.hpp
@@ -115,39 +115,10 @@ PORTABLE_INLINE_FUNCTION void printf(char const *const format, Ts... ts) {
 #endif // __HIPCC__
   return;
 }
-// Variadic strlen
-template <typename Head, typename... Tail>
-PORTABLE_INLINE_FUNCTION std::size_t strlen(Head h, Tail... tail) {
-  constexpr std::size_t MAX_I = 4096;
-  std::size_t i = 0;
-  if constexpr (std::is_convertible_v<Head, std::string_view>) {
-    // don't want a non-terminating loop if there's now null
-    // character.
-    for (i = 0; i < MAX_I; ++i) {
-      if (h[i] == '\0') {
-        break;
-      }
-    }
-  } else {
-    i = 100; // some big number to account for things like %.14e
-  }
-  if constexpr (sizeof...(Tail) > 0) {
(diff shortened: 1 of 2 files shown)


ports-of-call/portability.hpp
```diff
@@ -112,6 +113,17 @@ PORTABLE_INLINE_FUNCTION void printf(char const *const format, Ts... ts) {
...
+template <typename... Ts>
+PORTABLE_INLINE_FUNCTION void snprintf(char *target, std::size_t size, char const *const format, Ts... ts) {
+#ifndef __HIPCC__
+  std::snprintf(target, size, format, ts...);
+#endif // __HIPCC__
+  return;
+}
+template <typename... Ts>
+PORTABLE_INLINE_FUNCTION void sprintf(char *target, char const *const format, Ts... ts) {
+  PortsOfCall::snprintf(target, PORTABLE_MAX_NUM_CHAR, format, ts...);
```
BrendanKKrueger, line 125: The whole point of the compiler pushing you to use `snprintf` instead of `sprintf` is that you can set the character count based on the buffer size and prevent overflow.  Most buffers won't be 2048 characters wide.  Do you actually need `sprintf`?  It might make sense to only provide `snprintf`.  Or, if you can figure out some clever approach with templates or arguments or whatever, maybe a slightly non-standard `sprintf` that's able to figure out the buffer length?

Yurlungur: > The whole point of the compiler pushing you to use snprintf instead of sprintf is that you can set the character count based on the buffer size and prevent overflow. 

I know. But I am annoyed by the compiler telling me which language features I can and cannot use. In our context there's no danger.

> Most buffers won't be 2048 characters wide.

Fair enough.

> Do you actually need sprintf?

Not really it's only used in the `PortsOfCall::ErrorMessage` function, which is simply a convenience tool. However, automatically figuring out *exactlly* the buffer length is a pain because it requires walking the string and knowing the length that the language will use for every possible formatting directive in it.

> Or, if you can figure out some clever approach with templates or arguments or whatever, maybe a slightly non-standard sprintf that's able to figure out the buffer length?

This is a good idea and the path I went down now. That said, I'm not *exactly* getting the length. Just a bound on it. Because, as discussed above, getting an exact length is actually very painful. How do you feel about this?

```C++
// Variadic strlen
template<typename Head, typename... Tail>
inline std::size_t strlen(Head h, Tail... tail) {
  constexpr size_t MAX_I = 4096;
  std::size_t i = 0;
  if constexpr (std::is_convertible_v<Head, std::string_view>) {
  for (i = 0; i < MAX_I; ++i) {
    if (h[i] == '\0') {
        break;
    }
  }
  } else {
    i = 20; // some big number to account for things like %.14e
  }
  if constexpr (sizeof...(Tail) > 0) {
    i += strlen(tail...);
  }
  return i;
}
template <typename... Ts>
PORTABLE_INLINE_FUNCTION void snprintf(char *target, std::size_t size,
                                       char const *const format, Ts... ts) {
#ifndef __HIPCC__
  std::snprintf(target, size, format, ts...);
#endif // __HIPCC__
  return;
}
template <typename... Ts>
PORTABLE_INLINE_FUNCTION void sprintf(char *target, char const *const format, Ts... ts) {
#ifndef __HIPCC__
  std::size_t size = PortsOfCall::strlen(format, ts...);
  std::snprintf(target, size, format, ts...);
#endif // __HIPCC__
  return;
}
```

BrendanKKrueger: I can understand not liking the compiler bossing you around.  If this is because of the elevated error flags, you can always play around with the flags to tell the compiler to stop bossing you around.

I was more worried about overflowing `target`.  So, for example, the `std::string` version of `error_msg` knows how big `output_message` is, so it can set the correct limit for `snprintf` to avoid overflowing the string buffer.

BrendanKKrueger: That's a fair point.  If a buffer gets screwed up, we're already into error handling so it just means our error messages will be truncated or some such.

If you want to roll back to a fixed width, I'll leave that up to you.

BrendanKKrueger (approved):

ports-of-call/portability.hpp
Yurlungur: Nah your point about changing the function signature is good. I just changed the signature.

PR #73, opened 2024-12-05
add robust utils header

Yurlungur:
## PR Summary

As discussed over email.

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.

### DIFF
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -128,6 +128,65 @@ Please note that none of these functions are thread or MPI aware. In a parallel 
 
 as appropriate.
 
+robust_utils.hpp
+^^^^^^^^^^^^^^^^^^^
+
+``robust_utils.hpp`` contains small utility functions for numerical
+robustness, especially around floating point numbers. The available
+functionality is contained in the namespace ``PortsOfCall::Robust`` and includes:
+
+* ``constexpr auto SMALL<T>()`` returns a small number of type ``T``.
+* ``constexpr auto EPS<T>()`` returns a value of type ``T`` close to machine epsilon.
+* ``constexpr auto min_exp_arg<T>()`` returns the smallest of type ``T`` safe value to pass into an exponent.
+* ``constexpr auto max_exp_exp_arg<T>()`` returns the max value of type ``T`` to pass into an exponent.
+* ``auto make_positive(const T val)`` makes the argument of type ``T`` positive.
+
+where here all functionality is templated on type ``T`` and marked
+with ``PORTABLE_INLINE_FUNCTION``. The default type ``T`` is always
+``Real``.
+
+The function
+
+.. code-block:: cpp
+
+  template <typename T>
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real make_bounded(const T val, const T vmin, const T vmax);
+
+bounds ``val`` between ``vmin`` and ``vmax``.
+
+The function
+
+.. code-block:: cpp
+
+  template <typename T>
+  PORTABLE_FORCEINLINE_FUNCTION int sgn(const T &val);
+
+returns the sign of a quantity.
+
+The function
+
+.. code-block:: cpp
+
+  template <typename A, typename B>
+  PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b)
+
+computes the ratio :math:`A/B` but in a way robust to divide by zero
+errors. If both :math:`A` and :math:`B` are zero, this function will
+return 0.
+
+The function
+
+.. code-block:: cpp
+
+  template <typename T>
+  PORTABLE_FORCEINLINE_FUNCTION T safe_arg_exp(const T &x)
+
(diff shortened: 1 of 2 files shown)


ports-of-call/robust_utils.hpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+template <typename T>
+PORTABLE_FORCEINLINE_FUNCTION int sgn(const T &val) {
+  return (T(0) <= val) - (val < T(0));
+}
+
+template <typename A, typename B>
+PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b) {
+  return a / (b + sgn(b) * SMALL<B>());
```
BrendanKKrueger, line 62: I assume you're trying to avoid division by zero?  This doesn't actually avoid it, it simply makes `b` slightly larger (slightly more positive or slightly less negative to be more precise), and it just means there's a different "magic" value that will land exactly at zero.

doc/sphinx/src/using.rst
```diff
@@ -128,6 +128,65 @@ Please note that none of these functions are thread or MPI aware. In a parallel
...
+robust_utils.hpp
+^^^^^^^^^^^^^^^^^^^
+
+``robust_utils.hpp`` contains small utility functions for numerical
+robustness, especially around floating point numbers. The available
+functionality is contained in the namespace ``PortsOfCall::Robust`` and includes:
+
+* ``constexpr auto SMALL<T>()`` returns a small number of type ``T``.
+* ``constexpr auto EPS<T>()`` returns a value of type ``T`` close to machine epsilon.
+* ``constexpr auto min_exp_arg<T>()`` returns the smallest of type ``T`` safe value to pass into an exponent.
```
BrendanKKrueger, line 140: "... the smallest safe value of type ``T`` to pass..." would read a little better (same on the next line)

doc/sphinx/src/using.rst
```diff
@@ -128,6 +128,65 @@ Please note that none of these functions are thread or MPI aware. In a parallel
...
+
+The function
+
+.. code-block:: cpp
+
+  template <typename T>
+  PORTABLE_FORCEINLINE_FUNCTION
+  Real make_bounded(const T val, const T vmin, const T vmax);
+
+bounds ``val`` between ``vmin`` and ``vmax``.
```
BrendanKKrueger, line 156: It's probably worth noting that it's exclusive of the bounds themselves

doc/sphinx/src/using.rst
```diff
@@ -128,6 +128,65 @@ Please note that none of these functions are thread or MPI aware. In a parallel
...
+
+The function
+
+.. code-block:: cpp
+
+  template <typename T>
+  PORTABLE_FORCEINLINE_FUNCTION int sgn(const T &val);
+
+returns the sign of a quantity.
```
BrendanKKrueger, line 165: Is it worth noting that 0 is considered positive in the context of the `sgn` function?

doc/sphinx/src/using.rst
```diff
@@ -128,6 +128,65 @@ Please note that none of these functions are thread or MPI aware. In a parallel
...
+
+The function
+
+.. code-block:: cpp
+
+  template <typename A, typename B>
+  PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b)
+
+computes the ratio :math:`A/B` but in a way robust to divide by zero
```
BrendanKKrueger, line 174: See my comment on the code (it doesn't universally avoid division by zero), but also you should define what happens when you compute A / 0 -- Does it give infinity?  Does it give `std::numeric_limits::max`?  Does it use `small()` or `eps()` for B?

doc/sphinx/src/using.rst
BrendanKKrueger: Totally optional here, but it might also be interesting to note the similarities and differences to [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp).

Yurlungur: Yep will do.

doc/sphinx/src/using.rst
Yurlungur: Absolutely is. Will add a note.

doc/sphinx/src/using.rst
Yurlungur: As we discussed on mattermost, I made it a bit more robust with a masking operation. I also documented what happens when you have (finite)/0.

ports-of-call/robust_utils.hpp
Yurlungur: :+1: Good point! Fixed as we discussed on MM.

Yurlungur:
Thanks for the review, @BrendanKKrueger ! Feel free to push your changes to the same branch/MR. Or I can merge this and you can add yours in a separate MR. Up to you.

BrendanKKrueger (approved):

PR #74, opened 2024-12-09
Additional utilities

BrendanKKrueger:
## PR Summary

Adding some utilities from Singe that are also wanted for use in Nautilus or seemed appropriate to transfer to Ports-of-Call.

## PR Checklist

- [x] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


ports-of-call/robust_utils.hpp
```diff
@@ -64,13 +64,40 @@ PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b) {
...
+// type-safe check against 0
+template <typename T, std::enable_if_t<std::is_unsigned<T>::value, bool> = true>
+PORTABLE_FUNCTION constexpr bool check_nonnegative(const T) {
+  return true;
+}
+template <typename T, std::enable_if_t<!std::is_unsigned<T>::value, bool> = true>
+PORTABLE_FUNCTION constexpr bool check_nonnegative(const T t) {
+  return t >= 0;
+}
+
```
BrendanKKrueger, line 100: @Yurlungur and I discussed moving this out of `span.hpp`, because Nautilus and Singe both want to use it and I don't want to carry multiple implementations.

ports-of-call/robust_utils.hpp
```diff
@@ -64,13 +64,40 @@ PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b) {
...
+        return 1;
+    } else {
+        // Note: we do not use std::abs to calculate the absolute value because
+        //       it is not constexpr until C++23, therefore it is not portable
+        //       to GPUs (currently using C++17).
+        const auto signed_reldiff = (actual - expected) / expected;
+        return (signed_reldiff >= 0 ? signed_reldiff : -signed_reldiff);
+    }
+}
+
```
BrendanKKrueger, line 83: A simple relative difference calculation that is used extensively in the Singe unit tests and seemed a good candidate for here given that other codes probably need something similar for their unit tests.  If you feel this is out of scope, please raise an objection.

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,73 @@
...
+        exp >>= 1;
+        // If the exponent is now empty, we're done
+        if (!exp) {
+            break;
+        }
+        // We halved the exponent (see the shift operation), so square the base
+        base *= base;
+    }
+    return result;
+}
```
BrendanKKrueger, line 50: @dholladay00's `int_power` routine.  Currently it's in Singe, but I'm sure it appears in other libraries as well.  It seemed a likely candidate for Ports-of-Call, but feel free to object if it's out of scope.  I also wasn't sure where to put it, so I created a `math_utils.hpp` file.

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+template <typename IterB, typename IterE, typename Value, typename Op = singe::util::plus<Value>>
+PORTABLE_FUNCTION constexpr Value accumulate(
+    IterB begin, IterE end, Value accum, Op && op = singe::util::plus<Value>{})
+{
+    for (auto iter = begin; iter != end; ++iter) {
+        accum = op(accum, *iter);
+    }
+    return accum;
+}
```
BrendanKKrueger, line 68: Another thing from Singe that only exists because it's not `constexpr` in the STL.  I put it in the new `math_utils.hpp` header because it's kind of "math-y", although in the STL I think it may be in `algorithms`.  As with others: let me know if you have concerns about (a) this being out of scope, or (b) this being the wrong file for it.

ports-of-call/robust_utils.hpp
```diff
@@ -64,13 +64,40 @@ PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b) {
...
+  if (actual == expected) {
+    return 0;
+  } else if (expected == 0) {
+    // expected 0, got nonzero, define this to be 100% error
+    return 1;
+  } else {
+    // Note: we do not use std::abs to calculate the absolute value because
+    //       it is not constexpr until C++23, therefore it is not portable
+    //       to GPUs (currently using C++17).
+    const auto signed_reldiff = (actual - expected) / expected;
```
Yurlungur, line 79: Are you married to this particular form? I would prefer something like
```C++
Robust::ratio(2.*(actual - expected), actual + expected);
```
but if you need this exact form, that's fine.

ports-of-call/robust_utils.hpp
```diff
@@ -64,13 +64,40 @@ PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b) {
...
 }
 
+template <typename DataType>
+PORTABLE_FUNCTION constexpr DataType reldiff(const DataType actual,
+                                             const DataType expected) {
+  if (actual == expected) {
+    return 0;
+  } else if (expected == 0) {
+    // expected 0, got nonzero, define this to be 100% error
+    return 1;
```
Yurlungur, line 74: I think it's often useful to switch to an absolute error in this case, for example return `actual` here. But up to you.

doc/sphinx/src/using.rst
```diff
@@ -194,6 +194,73 @@ exceptions. For very large negative inputs, it returns 0. For very
...
+
+  template <typename DataType>
+  PORTABLE_FUNCTION constexpr DataType reldiff(const DataType actual, const DataType expected)
+
+returns the relative difference between the expected value and the actual
+result (:math:`\left|(A-E)/E\right|`).  If the expected and actual values are
+both zero, this is defined to be a relative difference of 0%.  If the expected
+value is zero and the actual value is non-zero, this is defined to be a
+relative difference of 100%.  Unlike ``ratio``, ``reldiff`` is tuned to get
+meaningful results for the use-case of unit tests, rather than for performance.
```
Yurlungur, line 209: As we discussed on MM, I found this an odd model, given we are working with floating point numbers. If this is a context where exact 0 can be expected, we should document that. Otherwise we might want to switch to something that switches to an absolute error if expected is exactly 0.

Either ways is fine, so long as this is named/documented to be what you need for singe. If we end up going with the latter case, then I can use this instead of the in-house function in singularity-eos.

Yurlungur (changes requested):
Looks good! Thanks for adding this! Only nitpick is: As we discussed offline, we should either document that `reldiff` is for things that might be exactly 0 or we should switch to it providing an absolute diff in that context. Whichever you pick is fine.

doc/sphinx/src/using.rst
BrendanKKrueger: I went back and looked at the usage of `reldiff` in Singe (which is where it came from).  It's pretty much exclusively (other than one or two printouts) used in the form `if(reldiff(actual, expected) <= tolerance)`.  In that case, we can trivially rewrite those clauses to be `if(abs(actual-expected) <= tolerance * abs(expected))`, which eliminates the entire divide-by-zero question.  Why I didn't do that earlier is unclear.

So what might be more useful than deciding on the "best" way to write `reldiff`, maybe we should just drop it from here and at some point I may change Singe to the other form.  But in that case, it _would_ be useful to add a GPU-compatible `abs` to `math_utils.hpp`.  What do you think about that plan?

ports-of-call/robust_utils.hpp
BrendanKKrueger: The form I used is more appropriate for "I know the right answer and want to compare against it" while your form is more appropriate for "I have two approximate answers and want to see how close they are".  But in practice, they both work well for these kinds of comparisons.  So it's not a huge deal either way.

That being said: see my comment about dropping `reldiff` entirely, because that would supersede this question.

doc/sphinx/src/using.rst
BrendanKKrueger: PS: I should say "eliminates the entire divide-by-zero question _and_ makes it explicit and under the user's control what to do if expected is zero".

I'm also hesitant to reproduce something like the Catch2 matchers for GPU, given my relatively simple use-cases.

Yurlungur: That's fine by me. We could alternatively implement something like the check you actually use in singe, sans the divide by zero.

Yurlungur: After our discussion, let's just punt on this and remove reldiff as you suggest. :+1:

Yurlungur (approved):
LGTM, thanks! @dholladay00 please check the int power implementation and let us know if you feel it's appropriate for `ports-of-call`. @BrendanKKrueger thought you might be concerned so we wanted to double check with you.

ports-of-call/math_utils.hpp
Yurlungur: I'm perfectly happy to include this. If we had more than one function to include, I'd suggest we add an algorithms header. But I'm not making you add a header file for one function. I say we leave it here for now. Maybe add a `TODO` mentioning that if we add more of this kind of thing, we should probably move it.

ports-of-call/math_utils.hpp
Yurlungur: This makes sense in `math_utils.hpp` and I'm perfectly happy to include it. We use it in `singularity-eos` too.

dholladay00:
Diego did more sophisticated work with power that may be useful here. I'd suggest reaching out to him to see if his work should be incorporated in this PR. I say that mainly b/c the names may change due to there being a global power function used and switched based on arg type (incorporating now would not incur an API change).

Lona85:
@dholladay00 @Yurlungur @BrendanKKrueger I can share my `power` implementation (used in EAP) if anybody is interested.

BrendanKKrueger:
> @dholladay00 @Yurlungur @BrendanKKrueger I can share my `power` implementation (used in EAP) if anybody is interested.

Yes, please do!

Yurlungur:
Ping @Lona85 can you share your power implementation? Would be great to get this in.

Lona85:
@BrendanKKrueger @Yurlungur @dholladay00 @junghans  It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?

Yurlungur:
> @BrendanKKrueger @Yurlungur @dholladay00 @junghans It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?

@Lona85 if you're comfortable doing so, you could push directly to @BrendanKKrueger 's branch. I can add you as a developer on the repository. Otherwise posting inline is fine. Up to you.

Lona85:
> > @BrendanKKrueger @Yurlungur @dholladay00 @junghans It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?
> 
> @Lona85 if you're comfortable doing so, you could push directly to @BrendanKKrueger 's branch. I can add you as a developer on the repository. Otherwise posting inline is fine. Up to you.

I pushed my `power.hpp` and `test_power.cpp` to @BrendanKKrueger 's branch. With some minimal modifications they should be ready to go.

junghans:
> > > @BrendanKKrueger @Yurlungur @dholladay00 @junghans It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?
> > 
> > 
> > @Lona85 if you're comfortable doing so, you could push directly to @BrendanKKrueger 's branch. I can add you as a developer on the repository. Otherwise posting inline is fine. Up to you.
> 
> I pushed my `power.hpp` and `test_power.cpp` to @BrendanKKrueger 's branch. With some minimal modifications they should be ready to go.

The push worked, now we just have to fix the compile:
```
[ 88%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_stats.cpp.o
In file included from /home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/internal/catch_enum_values_registry.cpp:9:
/home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/../catch2/internal/catch_string_manip.hpp:47:14: error: ‘uint64_t’ in namespace ‘std’ does not name a type; did you mean ‘wint_t’?
   47 |         std::uint64_t m_count;
      |              ^~~~~~~~
      |              wint_t
```

dholladay00:
> > > > @BrendanKKrueger @Yurlungur @dholladay00 @junghans It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?
> > > 
> > > 
> > > @Lona85 if you're comfortable doing so, you could push directly to @BrendanKKrueger 's branch. I can add you as a developer on the repository. Otherwise posting inline is fine. Up to you.
> > 
> > 
> > I pushed my `power.hpp` and `test_power.cpp` to @BrendanKKrueger 's branch. With some minimal modifications they should be ready to go.
> 
> The push worked, now we just have to fix the compile:
> 
> ```
> [ 88%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_stats.cpp.o
> In file included from /home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/internal/catch_enum_values_registry.cpp:9:
> /home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/../catch2/internal/catch_string_manip.hpp:47:14: error: ‘uint64_t’ in namespace ‘std’ does not name a type; did you mean ‘wint_t’?
>    47 |         std::uint64_t m_count;
>       |              ^~~~~~~~
>       |              wint_t
> ```

is this a missing header issue in catch2? Is there a patch or something?

Lona85:
> > > > @BrendanKKrueger @Yurlungur @dholladay00 @junghans It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?
> > > 
> > > 
> > > @Lona85 if you're comfortable doing so, you could push directly to @BrendanKKrueger 's branch. I can add you as a developer on the repository. Otherwise posting inline is fine. Up to you.
> > 
> > 
> > I pushed my `power.hpp` and `test_power.cpp` to @BrendanKKrueger 's branch. With some minimal modifications they should be ready to go.
> 
> The push worked, now we just have to fix the compile:
> 
> ```
> [ 88%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_stats.cpp.o
> In file included from /home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/internal/catch_enum_values_registry.cpp:9:
> /home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/../catch2/internal/catch_string_manip.hpp:47:14: error: ‘uint64_t’ in namespace ‘std’ does not name a type; did you mean ‘wint_t’?
>    47 |         std::uint64_t m_count;
>       |              ^~~~~~~~
>       |              wint_t
> ```

Note: Some implementation details are EAP-centric. The unit tests are in GoogleTest and not Catch2. I told @BrendanKKrueger to keep/adapt what he thinks might be useful for `ports-of-call` and discard the rest 😄

jonahm-LANL:
This is due to GitHub changing the default runner to Ubuntu 24.04. Changing the catch version to v3.7.1 in the cmake should resolve the issue

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Daniel Holladay ***@***.***>
Sent: Wednesday, December 18, 2024 3:27:28 PM
To: lanl/ports-of-call ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [lanl/ports-of-call] Additional utilities (PR #74)


@BrendanKKrueger<https://urldefense.com/v3/__https://github.com/BrendanKKrueger__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWglSbRRs$> @Yurlungur<https://urldefense.com/v3/__https://github.com/Yurlungur__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWgGAzvMu$> @dholladay00<https://urldefense.com/v3/__https://github.com/dholladay00__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWhlbKQuQ$> @junghans<https://urldefense.com/v3/__https://github.com/junghans__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWllPZSx-$> It does not look like I can attach my C++ files here. Should I just copy the source code in a comment?

@Lona85<https://urldefense.com/v3/__https://github.com/Lona85__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWqX09BMD$> if you're comfortable doing so, you could push directly to @BrendanKKrueger<https://urldefense.com/v3/__https://github.com/BrendanKKrueger__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWglSbRRs$> 's branch. I can add you as a developer on the repository. Otherwise posting inline is fine. Up to you.

I pushed my power.hpp and test_power.cpp to @BrendanKKrueger<https://urldefense.com/v3/__https://github.com/BrendanKKrueger__;!!Bt8fGhp8LhKGRg!FzCg5ETJ7X9U-WkdcSRi8h3sZCiYFdj8VlETnO8oIwA0ffujHKNPVC1K6XAEVsZ9nMuwswva8VcAUMMCWglSbRRs$> 's branch. With some minimal modifications they should be ready to go.

The push worked, now we just have to fix the compile:

[ 88%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_stats.cpp.o
In file included from /home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/internal/catch_enum_values_registry.cpp:9:
/home/runner/work/ports-of-call/ports-of-call/build/_deps/catch2-src/src/catch2/../catch2/internal/catch_string_manip.hpp:47:14: error: ‘uint64_t’ in namespace ‘std’ does not name a type; did you mean ‘wint_t’?
   47 |         std::uint64_t m_count;
      |              ^~~~~~~~
      |              wint_t


is this a missing header issue in catch2? Is there a patch or something?

BrendanKKrueger:
Flipping this back to draft while I massage Diego's code to fit within the Ports-of-Call framework/style (because I may not finish before leaving for the day).

BrendanKKrueger:
> This is due to GitHub changing the default runner to Ubuntu 24.04. Changing the catch version to v3.7.1 in the cmake should resolve the issue

Entirely possible that I did this wrong, but what I tried didn't work.

Yurlungur:
> > This is due to GitHub changing the default runner to Ubuntu 24.04. Changing the catch version to v3.7.1 in the cmake should resolve the issue
> 
> Entirely possible that I did this wrong, but what I tried didn't work.

This is on me---the github runner doesn't use pre-installed catch, it falls back to fetch content. Just pushed an update.

test/test_math_utils.cpp
```diff
@@ -0,0 +1,66 @@
+#include "ports-of-call/math_utils.hh"
```
Yurlungur, line 1:
```suggestion
#include <ports-of-call/math_utils.hpp>
```

BrendanKKrueger: Thanks!

BrendanKKrueger:
@Lona85 @Yurlungur Ready for review.  I think I've got Diego's power function(s) massaged to better fit within Ports-of-Call.  I squashed that down to a single MR to hide the fiddling.  Feel free to ask me to roll back some of those changes.

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+// Faster implementation of std::pow for arithmetic bases and non-negative integer
+// exponents.  For sufficiently large integer powers std::pow may be faster, but testing
+// indicates that the following implementation is significantly faster (roughly a factor
+// of two or better) up to powers of at least 100.
+template <typename BaseT, typename ExponentT,
+          typename std::enable_if<std::is_arithmetic_v<std::decay_t<BaseT>> &&
+                                  std::is_integral_v<std::decay_t<ExponentT>>>::type * =
+              nullptr>
+[[gnu::always_inline]] PORTABLE_INLINE_FUNCTION constexpr auto power(BaseT base,
```
Yurlungur, line 34: I suggest `PORTABLE_FORCEINLINE_FUNCTION` rather than the gnu compiler directive

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+  return result;
+}
+// Faster implementation of std::pow() for non-negative arithmetic bases and
+// floating-point exponents
+template <typename BaseT, typename ExponentT,
+          typename std::enable_if<std::is_arithmetic_v<std::decay_t<BaseT>> &&
+                                  std::is_floating_point_v<std::decay_t<ExponentT>>>::type
+              * = nullptr>
+[[gnu::always_inline]] PORTABLE_INLINE_FUNCTION constexpr auto
```
Yurlungur, line 56: As above, use `PORTABLE_FORCEINLINE_FUNCTION`

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+  return exponent == ExponentT{0} ? BaseT{1} // Enforcing base^0=1 (including 0^0=1)
+         : base == BaseT{0}       ? BaseT{0}
+                                  : exp(exponent * log(base));
+}
+// Overload for non-arithmetic bases or exponents
+template <typename BaseT, typename ExponentT,
+          typename std::enable_if<not std::is_arithmetic_v<std::decay_t<BaseT>> ||
+                                  not std::is_arithmetic_v<std::decay_t<ExponentT>>>::type
+              * = nullptr>
+[[gnu::always_inline]] PORTABLE_INLINE_FUNCTION constexpr auto
```
Yurlungur, line 73: `PORTABLE_FORCEINLINE`

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+power(BaseT const &base, ExponentT const &exponent) {
+  using std::pow;
+  return pow(base, exponent);
```
Yurlungur, line 76: seems a bit silly to use a `using` statement for a one-liner
```suggestion
  return std::pow(base, exponent);
```

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,99 @@
...
+  using std::exp;
+  using std::log;
+  using std::pow;
+  if (!Robust::check_nonnegative(base)) {
+    return pow(base, exponent);
+  }
+  return exponent == ExponentT{0} ? BaseT{1} // Enforcing base^0=1 (including 0^0=1)
+         : base == BaseT{0}       ? BaseT{0}
+                                  : exp(exponent * log(base));
+}
```
Yurlungur, line 67: These usings I don't think really help with readability given the simplicity of the function. I would prefer
```C++
power(BaseT const &base, ExponentT const &exponent) {
  if (!Robust::check_nonnegative(base)) {
    return std::pow(base, exponent);
  }
  return exponent == ExponentT{0} ? BaseT{1} // Enforcing base^0=1 (including 0^0=1)
         : base == BaseT{0}       ? BaseT{0}
                                  : std::exp(exponent * std::log(base));
}
```

Yurlungur (approved):
LGTM. Thanks for adding/cleaning this up, and fighting with the CI. A few nitpicks below. All are non-blocking.

ports-of-call/math_utils.hpp
BrendanKKrueger: Having worked a bunch with CJ, I'm pretty sure this came from him and is designed for ADL.  I had the same thought as you, but wasn't sure if we wanted to kill ADL.  Given that this will pretty much always be used for things like `int`, `double`, and other built-in POD types, it's probably a non-issue to drop ADL.  Having said that: Do you still prefer I make the recommended change?  If so, there are similar things elsewhere, should I change them too?

ports-of-call/math_utils.hpp
BrendanKKrueger: Same as above: Not about readability, but about ADL.  I'll let you decide if you still want the `using` statements gone.

Yurlungur: :+1: let's discuss in the above one

ports-of-call/math_utils.hpp
Yurlungur: I don't feel that strongly about it. If you're concerned we'll need non-standard types downstream for, e.g., this function to be called in EAP or Singe, we can leave it in.

Personally, I find aggressively type agnostic coding to be a big hit to readability and prefer being more explicit. But this is a shared capability and we can make compromises, especially since these functions in particular were obviously designed to be flexible as far as input type.

Yurlungur: Of all my comments, the thing I feel most strongly about is `PORTABLE_FORCEINLINE_FUNCTION` as that should be more flexible than manually inserting compiler specific directives about inlining.

BrendanKKrueger: I was happy to make the `PORTABLE_FORCEINLINE_FUNCTION` change.  I just forgot that existed and was copying code from somewhere else.

The change regarding the `using` statements has been made.

Yurlungur: Sounds good!

ports-of-call/math_utils.hpp
```diff
@@ -0,0 +1,94 @@
...
+// indicates that the following implementation is significantly faster (roughly a factor
+// of two or better) up to powers of at least 100.
+template <typename BaseT, typename ExponentT,
+          typename std::enable_if<std::is_arithmetic_v<std::decay_t<BaseT>> &&
+                                  std::is_integral_v<std::decay_t<ExponentT>>>::type * =
+              nullptr>
+PORTABLE_FORCEINLINE_FUNCTION constexpr auto power(BaseT base,
+                                                                     ExponentT exponent) {
+  using PowT = decltype(std::pow(base, exponent));
+  if (!Robust::check_nonnegative(exponent) || exponent > ExponentT{100}) {
```
dholladay00, line 37: I think there may be an additional optimization for negative exponents where you do `base = 1 / base;` and `exponent = abs(exponent);` and not have to fall back on the slow/robust pow. I know this PR is merged so perhaps something to do in the future @Lona85 @Yurlungur

Yurlungur: Very good suggestion. Thanks @dholladay00 !

PR #77, opened 2024-12-24
add bracket operator

Yurlungur:
## PR Checklist

Add the bracket operator to `PortableMDArray` as suggested by @mauneyc-LANL . This is a bridge contribution in support of #39 . Down the line, we may deprecate the 1D punned `operator()`.

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.
- [x] If preparing for a new release, update the version in cmake.

### DIFF
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -233,11 +233,16 @@ You can then set or access an element by reference as:
   // z = 3, y = 2, x = 1
   my_3d_array(3,2,1) = 5.0;
 
-You can always access the "flat" array by simply using the 1D accessor:
+You can always access the "flat" array by simply using the 1D bracket operator:
 
 .. code-block:: cpp
 
-  my_3d_array(6) = 2.0;
+  my_3d_array[6] = 2.0;
+
+.. warning::
+
+  Currently, the 1D parentheses operator currently also accesses the
+  flat array. However, this syntax may eventually be deprecated.
 
 By default ``PortableMDArray`` has reference-semantics. In
 other words, copies are shallow.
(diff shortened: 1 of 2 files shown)


ports-of-call/portable_arrays.hpp
```diff
@@ -157,6 +157,14 @@ class PortableMDArray {
...
   PORTABLE_FORCEINLINE_FUNCTION T *end() { return pdata_ + GetSize(); }
 
+  // 1D accessors intended for accessing flattened data
+  PORTABLE_FORCEINLINE_FUNCTION T &operator[](const int n) const {
+    return pdata_[n];
+  }
+  PORTABLE_FORCEINLINE_FUNCTION T &operator[](const int n) {
+    return pdata_[n];
+  }
+
```
BrendanKKrueger, line 167: Is this redundant if the two variations of the function do the same thing?  (I know I should know the answer to this, but it's first thing in the morning after a 2+ week vacation, so I'm a little slow...)

doc/sphinx/src/using.rst
```diff
@@ -233,11 +233,16 @@ You can then set or access an element by reference as:
...
 
-You can always access the "flat" array by simply using the 1D accessor:
+You can always access the "flat" array by simply using the 1D bracket operator:
 
 .. code-block:: cpp
 
-  my_3d_array(6) = 2.0;
+  my_3d_array[6] = 2.0;
+
+.. warning::
```
BrendanKKrueger, line 242: Do you want to use `deprecated` here or wait until it's _actually_ deprecated?

BrendanKKrueger (approved):

ports-of-call/portable_arrays.hpp
Yurlungur: Ah I think I made a mistake here. The `const` version should return a `const` reference. They should have different return types.

doc/sphinx/src/using.rst
Yurlungur: I would rather wait.

PR #83, opened 2025-03-04
Add PORTABLE_FENCE

jhp-lanl:
## PR Summary

Adds a `PORTABLE_FENCE` macro

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.

### DIFF
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -22,6 +22,7 @@ To include Ports of Call in your project, simply include the directory
 6. ``_WITH_CUDA_``: Defined when Cuda is enabled
 7. ``Real``: a typedef to double (default) or float (if you define ``SINGLE_PRECISION_ENABLED``)
 8. ``PORTABLE_MALLOC()``, ``PORTABLE_FREE()``: A wrapper for kokkos_malloc or cudaMalloc, or raw malloc and equivalent free.
+9. ``PORTABLE_FENCE()``: A wrapper for kokkos::fence or cudaDeviceSynchronize()
 
 At compile time, you define
 ``PORTABILITY_STRATEGY_{KOKKOS,CUDA,NONE}`` (if you don't define it,
@@ -71,7 +72,9 @@ limited. The syntax is:
     Function function, T &reduced) {
 
 where ``Function`` now takes as many indices are required and
-``reduced`` as arguments.
+``reduced`` as arguments. Note that a ``portableReduce()`` is blocking (i.e. a
+device synchronization step is performed) while `portableFor()` may not be,
+possibly requiring a ``PORTABLE_FENCE`` to avoid any race conditions.
 
 Also provided are host to device and device to host memory transfers of the form:
 
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -22,7 +22,7 @@ To include Ports of Call in your project, simply include the directory
 6. ``_WITH_CUDA_``: Defined when Cuda is enabled
 7. ``Real``: a typedef to double (default) or float (if you define ``SINGLE_PRECISION_ENABLED``)
 8. ``PORTABLE_MALLOC()``, ``PORTABLE_FREE()``: A wrapper for kokkos_malloc or cudaMalloc, or raw malloc and equivalent free.
-9. ``PORTABLE_FENCE()``: A wrapper for kokkos::fence or cudaDeviceSynchronize()
+9. ``PORTABLE_FENCE()``: A wrapper for ``kokkos::fence`` or ``cudaDeviceSynchronize()``
 
 At compile time, you define
 ``PORTABILITY_STRATEGY_{KOKKOS,CUDA,NONE}`` (if you don't define it,
--- a/ports-of-call/math_utils.hpp
+++ b/ports-of-call/math_utils.hpp
@@ -31,8 +31,7 @@ template <typename BaseT, typename ExponentT,
           typename std::enable_if<std::is_arithmetic_v<std::decay_t<BaseT>> &&
                                   std::is_integral_v<std::decay_t<ExponentT>>>::type * =
               nullptr>
-PORTABLE_FORCEINLINE_FUNCTION constexpr auto power(BaseT base,
-                                                                     ExponentT exponent) {
+PORTABLE_FORCEINLINE_FUNCTION constexpr auto power(BaseT base, ExponentT exponent) {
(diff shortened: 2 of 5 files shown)


jhp-lanl:
I don't think the `Check Formatting` step is working. I just ran clang format and modified some other files.

jhp-lanl:
Tests are running on our GPU machines

ports-of-call/math_utils.hpp
```diff
@@ -31,8 +31,7 @@ template <typename BaseT, typename ExponentT,
           typename std::enable_if<std::is_arithmetic_v<std::decay_t<BaseT>> &&
                                   std::is_integral_v<std::decay_t<ExponentT>>>::type * =
               nullptr>
-PORTABLE_FORCEINLINE_FUNCTION constexpr auto power(BaseT base,
-                                                                     ExponentT exponent) {
+PORTABLE_FORCEINLINE_FUNCTION constexpr auto power(BaseT base, ExponentT exponent) {
```
jhp-lanl, line 34: These changes were apparently because clang format wasn't run (or maybe a different version?).

ports-of-call/portability.hpp
```diff
@@ -53,6 +53,7 @@
 #define PORTABLE_MALLOC(size) Kokkos::kokkos_malloc<>(size)
 #define PORTABLE_FREE(ptr) Kokkos::kokkos_free<>(ptr)
 // Do we want to include additional terms here (for memory spaces, etc.)?
+#define PORTABLE_FENCE(name) Kokkos::fence(name)
```
jhp-lanl, line 56: Do I have access to Kokkos execution spaces? If so, it would make more sense to fence the device execution space by itself.

ports-of-call/math_utils.hpp
BrendanKKrueger: I believe it -- I've had trouble getting all the clang-format stuff to agree on various places I was working on Ports-of-Call.  By the way, the format job is still failing.  I'm not sure why it says it passes, but if you open it up and look at the output you'll see that there are failures.

Yurlungur: The formatting is due to the CI for ports-of-call apparently not being pinned to a specific clang-format version. I pinned it. That said, at some point we should update to a newer version of clang format.

ports-of-call/portability.hpp
Yurlungur: We don't... I think adding execution spaces generically unfortunately moves us more towards mimicking additional kokkos functionality. Let's leave this as is for now but add an issue?

doc/sphinx/src/using.rst
```diff
@@ -22,6 +22,7 @@ To include Ports of Call in your project, simply include the directory
 6. ``_WITH_CUDA_``: Defined when Cuda is enabled
 7. ``Real``: a typedef to double (default) or float (if you define ``SINGLE_PRECISION_ENABLED``)
 8. ``PORTABLE_MALLOC()``, ``PORTABLE_FREE()``: A wrapper for kokkos_malloc or cudaMalloc, or raw malloc and equivalent free.
+9. ``PORTABLE_FENCE()``: A wrapper for kokkos::fence or cudaDeviceSynchronize()
```
Yurlungur, line 25:
```suggestion
9. ``PORTABLE_FENCE()``: A wrapper for ``kokkos::fence`` or ``cudaDeviceSynchronize()``
```

ports-of-call/portability.hpp
Yurlungur, line 56: So you don't always need a name:
```suggestion
#define PORTABLE_FENCE(...) Kokkos::fence(__VA_ARGS__)
```

ports-of-call/portability.hpp
```diff
@@ -69,6 +70,7 @@ void *PORTABLE_MALLOC(size_t size) {
   return devPtr;
 }
 void PORTABLE_FREE(void *ptr) { cudaError_t e = cudaFree(ptr); }
+#define PORTABLE_FENCE(name) cudaDeviceSynchronize()
```
Yurlungur, line 73:
```suggestion
#define PORTABLE_FENCE(...) cudaDeviceSynchronize()
```

ports-of-call/portability.hpp
```diff
@@ -80,6 +82,7 @@ void PORTABLE_FREE(void *ptr) { cudaError_t e = cudaFree(ptr); }
 #define PORTABLE_LAMBDA [=]
 #define PORTABLE_MALLOC(size) malloc(size)
 #define PORTABLE_FREE(ptr) free(ptr)
+#define PORTABLE_FENCE(name)
```
Yurlungur, line 85:
```suggestion
#define PORTABLE_FENCE(...)
```

test/test_portability.cpp
```diff
@@ -132,3 +132,58 @@ TEST_CASE("portableCopy works with all portability strategies", "[portableCopy]"
...
+  portableReduce(
+      "Check PORTABLE_FENCE", 0, N,
+      PORTABLE_LAMBDA(const int &i, int &n_wrong) {
+        if (d_data[i] != init_val * mult_factor) {
+          n_wrong += 1;
+        }
+      },
+      n_wrong);
+
+  REQUIRE(n_wrong == 0);
```
Yurlungur, line 185: I can't think of anything better, but I'm actually not sure this would necessarily fail if the fence wasn't in place... I *think* multiple `Kokkos::parallel_for`s in the same execution space somehow execute one-after-the-other. I have never fully understood the rules there.

Yurlungur (approved):
Some nitpicks below. Formatting should pass now.

ports-of-call/portability.hpp
jhp-lanl: 🤷 I'm not sure it even needs an issue since, as you said, that's going a bit beyond what we want to do.

test/test_portability.cpp
jhp-lanl: Yeaaaaaah I'm not sure. All I could discover was that there is an implicit fence at the end of a `parallel_reduce`, but that there _shouldn't_ be one at the end of `parallel_for`. I don't know what that means for what I've written though.

jhp-lanl: Changed. I also increased the number of array items in case the first loop can continue on that task while the reduction goes in. The fence will ensure the answer doesn't change, but if somebody wants to test this, they can remove the fence and see what happens.

PR #84, opened 2025-05-23
update clang format to 19 and perform format sweep

Yurlungur:
## PR Summary

Again, just doing this to make formatting easier. Time to move on from super outdated clang-format-12.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/cmake/Format.cmake
+++ b/cmake/Format.cmake
@@ -1,5 +1,5 @@
 #=========================================================================================
-# (C) (or copyright) 2024. Triad National Security, LLC. All rights reserved.
+# (C) (or copyright) 2025. Triad National Security, LLC. All rights reserved.
 #
 # This program was produced under U.S. Government contract 89233218CNA000001 for Los
 # Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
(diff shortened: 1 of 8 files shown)


cmake/Format.cmake
```diff
@@ -1,5 +1,5 @@
 #=========================================================================================
-# (C) (or copyright) 2024. Triad National Security, LLC. All rights reserved.
+# (C) (or copyright) 2025. Triad National Security, LLC. All rights reserved.
```
BrendanKKrueger, line 2: This is one of those comments that you're welcome to ignore if you like, and would probably be best in another MR anyway, but: the Feynman Center has agreed that copyright statements of this form can be in a separate copyright file within your repository.  Having a copyright statement in every file almost guarantees that the copyright statements will be inconsistent with each other, and typically will be outright wrong in several cases after a while.  Instead of updating them like this, you might consider deleting them all and adding a copyright file.

script/format.sh
```diff
@@ -27,7 +27,7 @@ else
 fi
 
 # clang format major version
-TARGET_CF_VRSN=12
+TARGET_CF_VRSN=19
```
BrendanKKrueger, line 30: The number of places you have to update from 12 to 19 seems rather high.  I'm not sure if it's possible to streamline things (a CMake variable or environment variable or some such?), but it might be helpful for future changes like this.

BrendanKKrueger (approved):

PR #87, opened 2025-07-22
Prepare for Spack v1.0.0

rbberger:
## PR Summary

These changes are backwards compatible all the way to v0.23. Since Singularity-EOS already has them, we should just bring this library up-to-date. There is more Spack related changes incoming due to the Package API v2 changes, but that will be its own PR.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.


Yurlungur (approved):

jhp-lanl:
thanks for being so forward-looking @rbberger !

PR #94, opened 2025-11-19
Change CMake install location to be relative to ports-of-call

alexrlongne:
## PR Summary

* Change `CMAKE_SOURCE_DIR` to `PROJECT_SOURCE_DIR` to get a path relative to ports-of-call CMake project instead of relative to top level CMake source (enables installs where ports of call is fetched and built by cmake). In our project ports-of-call thought its headers were located in our source path instead of the fetched ports-of-call source tree. This should allow installations where ports-of-call is fetched and built as a dependency.

* Old CMake variable: `CMAKE_SOURCE_DIR`: https://cmake.org/cmake/help/latest/variable/CMAKE_SOURCE_DIR.html
* New CMake variable: `PROJECT_SOURCE_DIR`: https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${POCLIB}Config.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${POCLIB})
 
 # Install the source header files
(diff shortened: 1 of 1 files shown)


rbberger (approved):
LGTM

PR #109, opened 2026-04-06
Expose execution space in portable dispatch

Yurlungur:
## PR Summary

This change is motivated by two issues in singularity-eos:
1. @rbberger indicates that Flecsi will at some point want to be able to choose how `portableFor` calls and the singularity-eos vector API calls are launched to launch on host or device, depending on context.
2. In riot we are hitting this right now with Python. In particular, we want to build for a GPU system but enable the Python API to function with CPU calls. https://github.com/lanl/singularity-eos/issues/628

As a first step towards fixing that, I thread through `ports-of-call` portable functions the ability to select an execution space. It's really only implemented for Kokkos and serial backends at the moment. @adamdempsey90 this probably also relates to your memory pool MR #105 .

I used codex to eliminate some of the drudgery of this MR. Which forced me to think through contribution rules for agentic-contributed code. Those rules are also in this MR. Feedback welcome.

## PR Checklist

- [x] Code is formatted.
- [x] Install test passes.
- [x] Docs build.
- [x] If ML was used, make sure to add a disclaimer at the top of a file indicating ML was used to assist in generating the file.
- [x] If Agentic AI was used, have the AI generate a "proposed changes" markdown file and store it in the `plan_histories` folder, with a filename the same as the MR number.

### DIFF
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -449,3 +449,7 @@ not be thrown on device.
 
 .. _`mpark variant`:  https://github.com/mpark/variant
 .. _`variant`: https://en.cppreference.com/w/cpp/utility/variant.html
+
+.. note::
+
+  This file was generated in part with generative AI.
--- a/doc/sphinx/src/using.rst
+++ b/doc/sphinx/src/using.rst
@@ -39,7 +39,7 @@ portability.hpp
 functions. Also provides loop abstractions that can be leveraged by a
 code. These loop abstractions are of the form:
 
-.. cpp:function:: void portableFor(const char *name, int start, int stop, Function Function)
+.. cpp:function:: template <PortsOfCall::Exec E = PortsOfCall::Exec::Device, typename Function> void portableFor(const char *name, int start, int stop, Function function)
 
 where ``Function`` is a template parameter and should be set to a
 functor that takes one index, e.g., an index in an array. For example:
@@ -51,12 +51,27 @@ functor that takes one index, e.g., an index in an array. For example:
       printf("hello from thread %d\n", i);
   });
 
+The optional template parameter ``E`` selects the execution space to
+launch on. The default is ``PortsOfCall::Exec::Device``, which preserves
+the historical behavior. Under Kokkos, ``PortsOfCall::Exec::Device``
+maps to ``Kokkos::DefaultExecutionSpace`` and ``PortsOfCall::Exec::Host``
+maps to ``Kokkos::DefaultHostExecutionSpace``.
+
+For example, to force a loop onto the host execution space:
+
+.. code-block:: cpp
+
+  portableFor<PortsOfCall::Exec::Host>("ExampleHost", 0, 5,
+    PORTABLE_LAMBDA(int i) {
+      printf("hello from host thread %d\n", i);
+  });
+
 ``start`` is inclusive, ``stop`` is exclusive. Up to five-dimensional
 ``portableFor`` loops are available. For example:
 
 .. code-block:: cpp
 
-  template <typename Function>
+  template <PortsOfCall::Exec E = PortsOfCall::Exec::Device, typename Function>
   void portableFor(const char *name, int startb, int stopb, int starta, int stopa,
     int startz, int stopz, int starty, int stopy, int startx,
     int stopx, Function function) {
@@ -66,7 +81,7 @@ limited. The syntax is:
 
 .. code-block::
 
-  template <typename Function, typename T>
+  template <PortsOfCall::Exec E = PortsOfCall::Exec::Device, typename Function, typename T>
   void portableReduce(const char *name, int starta, int stopa, int startz,
     int stopz, int starty, int stopy, int startx, int stopx,
     Function function, T &reduced) {
@@ -75,6 +90,24 @@ where ``Function`` now takes as many indices are required and
 ``reduced`` as arguments. Note that a ``portableReduce()`` is blocking (i.e. a
 device synchronization step is performed) while `portableFor()` may not be,
 possibly requiring a ``PORTABLE_FENCE`` to avoid any race conditions.
+
+As with ``portableFor``, the optional ``E`` template parameter can be
+used to select host or device execution explicitly. For example:
+
+.. code-block:: cpp
+
+  int sum = 0;
+  portableReduce<PortsOfCall::Exec::Host>(
+    "HostReduce", 0, 5,
+    PORTABLE_LAMBDA(int i, int &local_sum) {
+      local_sum += i;
+    }, sum);
+
+When selecting ``PortsOfCall::Exec::Host``, the lambda or functor and
+the memory it touches must be valid on host. When selecting
+``PortsOfCall::Exec::Device``, device-accessible storage should be used,
+for example memory returned by ``PORTABLE_MALLOC()`` under device
+backends.
 
 Also provided are host to device and device to host memory transfers of the form:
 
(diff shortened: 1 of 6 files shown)


rbberger:
@Yurlungur need to be a bit careful with the wording here. I never said anything about FleCSI in this context. I've mentioned Ristra as a possible client that wants to control whether vectorized calls happen on host or device. FleCSI isn't going to use any of this.

Yurlungur:
> @Yurlungur need to be a bit careful with the wording here. I never said anything about FleCSI in this context. I've mentioned Ristra as a possible client that wants to control whether vectorized calls happen on host or device. FleCSI isn't going to use any of this.

Ah thank you for the correction.

doc/sphinx/src/using.rst
```diff
@@ -76,6 +91,24 @@ where ``Function`` now takes as many indices are required and
...
+
+.. code-block:: cpp
+
+  int sum = 0;
+  portableReduce<PortsOfCall::Exec::Host>(
+    "HostReduce", 0, 5,
+    PORTABLE_LAMBDA(int i, int &local_sum) {
+      local_sum += i;
+    }, sum);
+
```
adamdempsey90: I kind of prefer the style of passing the execution space as an argument to the portableFor or whatever and making aliases to `DevExecSpace` and `HostExecSpace`. That follows what we've done in other code bases and I think that's worked out reasonable well?

ports-of-call/portability.hpp
adamdempsey90: Should we pass the execution space to `deep_copy` too? and should this be `Kokkos::deep_copy`? Similar question on the CopyToHost deep copy

adamdempsey90: Should we take this opportunity to make this not a macro but a forced inline function that takes an execution space's memory space for the malloc? Same comment for free

adamdempsey90 (approved):

doc/sphinx/src/using.rst
Yurlungur: Yeah we could do that. The reason I went with the template parameter is just it's a smaller diff, as the exact code can be re-used. Your version, I think ends up with a function overload for the optional argument, if we want to preserve backwards compatibility.  Happy to change it if you feel strongly.

ports-of-call/portability.hpp
Yurlungur: Moving this off of being a macro, I think, would ideally mean making it lower case... which would be an API breaking change. Or do you think it's fine to just leave it all caps?

Yurlungur: This definitely should be a `Kokkos::deep_copy`. Regarding the execution space, the current API is `portableCopyToHost` and `portableCopyToDevice`, which is pretty explicit. We could add a `portableCopy` that takes an execution space, though. I'll add that.

BrendanKKrueger: Introduce the function (no caps), leave the macro (all caps) as basically just an alias for the function, document that the macro is now deprecated and the new function is now preferred, at the next major release delete the macro.

Yurlungur: Done.

Yurlungur: Done.

doc/sphinx/src/using.rst
Yurlungur: Ok I can just support both very easily with variadics. This will get unwieldy if we expose more complex parallelism (we'd basically end up copying parthenon), so let's not do that. But this level of complexity is not bad at all.

Yurlungur: Ah, actually the version as it is now treats the type still as an enum rather than its own type. I'll fix that.

rbberger (commented):
This assumes we want to use the default host or device execution spaces. I think in the design of this you should consider the possibility of different execution spaces passed in from the outside. So even though you abstract this with your own type, it should allow different execution space instances being passed. Two example come to mind:
- multiple OpenMP execution spaces (can happen with partition_space)
- multiple CUDA/HIP streams, e.g. for targeting different GPUs

We have the same design problem in FleCSI, so coordination would make sense. @opensdh FYI

opensdh:
It was really a mistake that Kokkos ever allowed execution spaces to be specified as types devoid of values; libraries should just accept the values (thus deducing the types) and leave their selection/"creation" to the client.

ports-of-call/portability.hpp
```diff
@@ -170,12 +237,21 @@ void portableCopyToHost(T *const to, T const *const from, size_t const size_byte
...
 }
+template <typename T>
+void portableCopy(PortsOfCall::Exec e, T *const to, T const *const from,
+                  size_t const size_bytes) {
+  if (e == PortsOfCall::Exec::Device) {
+    portableCopyToDevice(to, from, size_bytes);
+  } else { // host
+    portableCopyToHost(to, from, size_bytes);
+  }
+}
```
adamdempsey90: I think you always want to pass device execution space even if it's device to host. That basically queues up the copy in the device's work queue and lets the host move on (i.e., it's async). I think it's only for host->host copy that you pass the host execution space.

Also, since passing the execution space makes it async, maybe that should be called out in the documentation

doc/sphinx/src/using.rst
```diff
@@ -21,8 +21,7 @@ To include Ports of Call in your project, simply include the directory
 5. ``_WITH_KOKKOS_``: Defined if Kokkos is enabled.
 6. ``_WITH_CUDA_``: Defined when Cuda is enabled
 7. ``Real``: a typedef to double (default) or float (if you define ``SINGLE_PRECISION_ENABLED``)
-8. ``PORTABLE_MALLOC()``, ``PORTABLE_FREE()``: A wrapper for kokkos_malloc or cudaMalloc, or raw malloc and equivalent free.
-9. ``PORTABLE_FENCE()``: A wrapper for ``kokkos::fence`` or ``cudaDeviceSynchronize()``
+8. ``PORTABLE_FENCE()``: A wrapper for ``kokkos::fence`` or ``cudaDeviceSynchronize()``
```
adamdempsey90, line 24: There is also the execution space fence (i.e., the "softer" fence). It might be worth giving this the PORTABLE_MALLOC treatment and make an inline version that takes the execution space and invokes its fence.

ports-of-call/portability.hpp
Yurlungur: I don't think I understand what you want here. Let's discuss offline and delay this change to a later MR.

doc/sphinx/src/using.rst
Yurlungur: I started to go down this path, but actually because of how it looks a bit difference than the global fence, I'd like to put this off for a later MR. I made an issue to tackle this later: #110 .

ports-of-call/portability.hpp
Yurlungur: Discussed offline. I added a todo but for now removing this.

doc/sphinx/src/using.rst
Yurlungur: Ok now execution spaces are transparently passed through and Kokkos spaces are available. Under the hood, it still selects via a template argument, not via an argument, but that is not documented and I added a comment saying we reserve the right to change it.

Yurlungur:
Thanks for the detailed review @adamdempsey90 . I think I've addressed all your comments. Feel free to re-review if desired. I think we discussed offline, so no need if you feel you're happy.

@rbberger @opensdh thanks for your comments/suggestions. I now pass types through transparently so Kokkos types are available and they are passed in as arguments. Hopefully that should be sufficiently compatible with ristra. Broadly, I am happy to iterate via future pull requests, but I would really like to keep this MR minimal and avoid feature creep. I am not trying to mimic or pass through all of Kokkos. `ports-of-call` is supposed to be a very thin, minimalistic thing.

adamdempsey90 (approved):
LGTM

Yurlungur:
It turned out to be pretty trivial to swap the overload logic so the space is always passed in to the range policy, so I made that change. Will now merge when tests pass on re-git.

Yurlungur:
Tests passing. Merging it. @rbberger @opensdh happy to iterate on this going forward. Hopefully the new version meets Ristra's needs, as execution space object is now passed in to the execution policies.

PR #112, opened 2026-05-08
add tests to robust utils, update to C++20, add Robust::is_normal

Yurlungur:
## PR Summary

This MR is a bit of a cleanup. I:
1. Update us to C++20 as we discussed in the call
2. Add tests for robust_utils, which were previously not present. This process unearthed some bugs, which are fixed.
3. Add `Robust::is_normal` and `Robust::is_normal_or_zero`, which I move out of [this MR in singularity-eos](https://github.com/lanl/singularity-eos/pull/637). The main motivation here is that `std::is_normal` is not portable. This provides a portable alternative. I also take the opportunity to start using concepts now that C++20 is supported. I strategically upgrade that where appropriate.

## PR Checklist

- [x] Code is formatted.
- [ ] Install test passes.
- [x] Docs build.
- [x] If ML was used, make sure to add a disclaimer at the top of a file indicating ML was used to assist in generating the file.
- [ ] If Agentic AI was used, have the AI generate a "proposed changes" markdown file and store it in the `plan_histories` folder, with a filename the same as the MR number.

### DIFF
--- a/ports-of-call/robust_utils.hpp
+++ b/ports-of-call/robust_utils.hpp
@@ -18,6 +18,7 @@
 #define PORTS_OF_CALL_ROBUST_UTILS_HPP_
 
 #include <algorithm>
+#include <concepts>
 #include <cmath>
 #include <limits>
 #include <ports-of-call/portability.hpp>
--- a/ports-of-call/robust_utils.hpp
+++ b/ports-of-call/robust_utils.hpp
@@ -25,9 +25,8 @@
 namespace PortsOfCall {
 namespace Robust {
 
-template <typename T = Real,
-          typename std::enable_if<std::is_integral<T>::value ||
-                                  std::is_floating_point<T>::value>::type * = nullptr>
+template <typename T = Real>
+  requires(std::integral<T> || std::floating_point<T>)
 PORTABLE_FORCEINLINE_FUNCTION constexpr bool is_normal(const T val,
                                                        const T factor = T{1}) {
   if constexpr (std::is_integral_v<T>) {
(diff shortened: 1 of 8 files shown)


ports-of-call/robust_utils.hpp
```diff
@@ -12,16 +12,41 @@
...
+  }
+}
+
+template <typename T = Real,
+          typename std::enable_if<std::is_integral<T>::value ||
+                                  std::is_floating_point<T>::value>::type * = nullptr>
+PORTABLE_FORCEINLINE_FUNCTION constexpr bool is_normal_or_zero(const T val,
+                                                               const T factor = T{1}) {
+  return (val == T{0}) || is_normal(val, factor);
+}
```
adamdempsey90, line 47: Your ordering implies the function name is `is_zero_or_normal` ;)

ports-of-call/robust_utils.hpp
```diff
@@ -59,9 +84,9 @@ PORTABLE_FORCEINLINE_FUNCTION int sgn(const T &val) {
 
 template <typename A, typename B>
 PORTABLE_FORCEINLINE_FUNCTION auto ratio(const A &a, const B &b) {
-  B mask = static_cast<B>(std::abs(b) < SMALL());
+  B mask = static_cast<B>(std::abs(b) < SMALL<B>());
   B denom = mask * sgn(b) * SMALL<B>() + (1 - mask) * b;
```
adamdempsey90:
```suggestion
const B mask = static_cast<B>(std::abs(b) < SMALL<B>());
const B denom = mask * sgn(b) * SMALL<B>() + (1 - mask) * b;
```

adamdempsey90 (approved):

Yurlungur:
I will do a subsequent sweep through the code to update more SFINAE to concepts

ports-of-call/robust_utils.hpp
Yurlungur: lol fair. zero commutes though :P

ports-of-call/span.hpp
```diff
@@ -26,25 +26,13 @@
...
-namespace PortsOfCall::span {
-
-using std::span;
-
-} // namespace PortsOfCall::span
-
-#else
-#ifdef PORTS_OF_CALL_USES_STD_SPAN
-#undef PORTS_OF_CALL_USES_STD_SPAN
-#endif
```
Yurlungur, line 42: Tests for span fail on GPU if we swap out for std::span. I would like to move to C++20 first then if std::span functions downstream, then we can remove span and tests for it.

PR #114, opened 2026-06-22
Remove span

mauneyc-LANL:
## PR Summary

With C++20 required, we no longer need to support `span`. 

Removed `PortsOfCall::span` and associated tests.

## PR Checklist

- [ ] Code is formatted.
- [ ] Install test passes.
- [ ] Docs build.
- [ ] If ML was used, make sure to add a disclaimer at the top of a file indicating ML was used to assist in generating the file.
- [ ] If Agentic AI was used, have the AI generate a "proposed changes" markdown file and store it in the `plan_histories` folder, with a filename the same as the MR number.


Yurlungur (approved):
Assuming @BrendanKKrueger is happy with `std::span` I'm happy to remove.

BrendanKKrueger (approved):
No objection to removing PoC's span in favor of `std::span`.  The weirdness with Singe + GPUs + span still isn't resolved, but that has no impact on Ports-of-Call since the same behavior is seen with PoC's span and `std::span`.

I'll have to check if Nautilus, Singe, or SuNRe is using PoC's span and update.  But that doesn't block this MR -- it just blocks updating to a new version of PoC.

Yurlungur:
:+1: @mauneyc-LANL are tests passing on re-git? If so, I'll press the button.

